PR #2, opened 2016-12-01
migrate master from Originate/exorelay-java

hugobho:
@kevgo 

I think I used some IDE to run the Cucumber tests, so I got rid of the `bin/spec` files. Unfortunately that causes Morula to crash because it can't find the tests.


exorelay-java/bin/setup
```diff
@@ -1,4 +0,0 @@
-#!/usr/bin/env bash
```
kevgo: Don't delete this file

exorelay-java/bin/spec
```diff
@@ -1,4 +0,0 @@
-#!/usr/bin/env bash
```
kevgo: Don't delete this file

exorelay-java/bin/spec_ci
```diff
@@ -1,4 +0,0 @@
-#!/usr/bin/env bash
```
kevgo: Don't delete this file

PR #6, opened 2016-12-08
Change `to-lower-case` to `lowercase-keys`

martinjaime:
Resolves Originate/exosphere-sdk#74

Uses `lowercase-keys` on an object instead of `to-lower-case` on each key.

@kevgo, @hugobho, @trushton, I feel that this issue has no positive impact since it doesn't make the code simpler, and it adds an extra module. The use of `to-lower-case` has similar calls in other exosphere related repos.


exocom-mock-javascript/features/steps/steps.ls
```diff
@@ -106,7 +107,7 @@ module.exports = ->
 
   @Then /^it has received the messages/, (table, done) ->
     wait-until (~> @exocom.received-messages.length > 1), 10, ~>
-      expected-messages = [{[key.to-lower-case!, value] for key, value of message} for message in table.hashes!]
+      expected-messages = [{[key, value] for key, value of lowercase-keys message} for message in table.hashes!]
```
kevgo: This can be further simplified to:

```ls
expected-messages = [lowercase-keys(message) for message in table.hashes!]

kevgo: or

```ls
table.hashes! |> map -> lowercase-keys it

martinjaime: cool, thanks!

PR #8, opened 2016-12-20
Automatically read enviroment variables

martinjaime:
So that the user can instantiate exorelay-javascript without arguments

resolves Originate/exosphere-sdk#37


martinjaime:
Not a good idea. Closed by @kevgo

PR #14, opened 2017-01-02
KG 5: specify the payload of registration messages

kevgo:
The messages with which services register at ExoCom should be specified by Cucumber, so that it is clear how to register with ExoCom. This PR implements this for both ExoCom and MockExocom.

@trushton @hugobho


exocom-server/features/support/cli-world.ls
```diff
@@ -27,7 +27,8 @@ CliWorld = !->
 
   @create-mock-service-at-port = ({name, port, namespace}, done) ->
     (@service-mocks or= {})[name] = new MockService {port, name, namespace}
-    wait 200, done
+    @service-mocks[name].connect {}, ->
+      wait 200, done
```
hugobho: You're working on a PR for exocom-server to reply with a message once a service has registered itself right? Does that mean this `wait` statement will instead be for that reply message in the future?

kevgo: In the future we will just wait for the reply from Exocom. The service-mock will signal that it received that message, for example by emitting an event. This way we wait here exactly until the environment is set up, no matter how fast or how slow that is. 

Exorelays and services will also wait for that message before they signal that they are online.

PR #16, opened 2017-01-02
KG 9: reorganize steps

kevgo:
This PR doesn't change anything, just organizes steps into given/when/then files, and sorts them alphabetically.

@hugobho @trushton


trushton:
@kevgo Should we use this same organization on the `exosphere` monorepo also?

kevgo:
Yes, I think that makes sense. I have been using it on a number of other repos, and it seems to help find steps faster.

PR #18, opened 2017-01-02
KG 8: remove multi-level testing for Exocom-Server

kevgo:
Exocom-server does not provide a JS API anymore, so this PR cleans it out.

@hugobho @trushton


exocom-server/bin/features
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+node_modules/o-tools-livescript/bin/build
+if [ "$#" == "0" ]; then
+  node_modules/.bin/cucumber-js --tags ~@todo
+else
+  DEBUG='*,-express:*,-body-parser:*,-finalhandler,-text-stream-search,-rails-delegate' node_modules/.bin/cucumber-js --tags ~@todo "$@"
```
hugobho: I don't think we have to test `-rails-delegate` here right?

PR #21, opened 2017-01-02
KG 12: clean up Exocom's HTTP subsystem

kevgo:
The HTTP subsystem had a bunch of technical drift like old names that no longer apply, and a separation into 2 files that no longer makes sense. This PR simplifies it to one file and improves the way it signals events to the Exocom core. 

@hugobho @trushton


exocom-server/src/exocom.ls
```diff
@@ -15,12 +15,11 @@ class ExoCom extends EventEmitter
...
   ({@service-messages} = {}) ->
     @client-registry    = new ClientRegistry {@service-messages}
-    @listener-subsystem = new ListenerSubsystem @
+    @http-subsystem = new HttpSubsystem @
+      ..on 'online', (port) ~> @emit 'http-online', port
     @message-cache = new MessageCache!
     @websocket = new WebSocketSubsystem @
 
-    delegate 'httpPort', from: @, to: @listener-subsystem
-    delegate-event 'http-bound' 'error', from: @listener-subsystem, to: @
```
hugobho: Shouldn't `error` from `@http-subsystem` still be bubbled up to `@`?

hugobho: Nevermind, I see that you've made those changes in another PR.

hugobho: Nevermind, I see you've made those changes in another PR.

kevgo: Good catch, that change probably slipped into the wrong PR!

PR #23, opened 2017-01-02
KG 14: clean up the Exocom CLI

kevgo:
- extract a logger class that encapsulates interaction with the CLI
- in the future, we can add other loggers that write JSON to a log file, or send log data to a third-party logging service
- the cli file now only boots up the Exocom core and hooks the logger to it

@trushton @hugobho


exocom-server/src/cli-logger.ls
```diff
@@ -0,0 +1,60 @@
...
+  write: (text) ~>
+    process.stdout.write text
+
+
+  log-header: (text) ->
+    @log dim text
+
+
+  log-websockets-online: (port) ~>
+    @log "ExoCom WebSocket listener online at port #{cyan port}"
```
kevgo: Breaking change: This no longer prints the Exocom version. We print it at the beginning, so it was not needed here and just made waiting for this output unnecessarily difficult.

exocom-server/src/cli-logger.ls
```diff
@@ -0,0 +1,60 @@
...
+
+
+  log-http-online: (port) ~>
+    @log "ExoCom HTTP service online at port #{magenta port}"
```
kevgo: Same breaking change here

PR #26, opened 2017-01-04
KG 17: rename "name" to "service type"

kevgo:
__BREAKING CHANGE:__ changes the format of the environment variables via which we tell Exocom about services when booting it up

Over time we have started using "name" for a number of different concepts, including the name of the service class and the name of the service instance. This makes the code confusing, because it's not always clear what exact piece of data one is dealing with. I am going to clean this up by introducing more correct names for a number of concepts:
- __service type__ denotes the class of service "web", "users service", etc
- __client name__ denotes the name of an instance of a service type. If we have 2 instances of the "web" service type, they could be named "web 1" and "web 2". I'm picking "client name" because "service instance name" is too long, and I don't want to use "service again", and service instances can be considered clients of Exocom.

Thoughts and feedback appreciated.

This PR changes "name" to "service type"

@trushton @hugobho


hugobho:
Per our discussion during today's standup, "service name" should actually be renamed to "service role". See: https://github.com/Originate/exocom/issues/35

kevgo:
I have changed "serviceType" to "role", according to our conversation at the standup.

@hugobho I think the position you point to refers to where service instances (aka clients) register themselves with the websocket subsystem. The websocket subsystem doesn't care what role instances have. It just keeps a list of which instance has which websocket. So that when Exocom wants to send something to an instance, the websocket subsystem does that job for it.

kevgo:
@hugobho you are right, I think we need to rename the env variable SERVICE_NAME to ROLE when starting exoservices and exorelays.

PR #27, opened 2017-01-04
KG 18: improve parameter name

kevgo:
__BREAKING CHANGE:__ changes the name of the environment variable that configures service routes from SERVICE_MESSAGES to SERVICE_ROUTES

_Messages_ are actual messages sent between services. _Routes_ is the information which _service type_ can send which _message type_.

@trushton @hugobho

### DIFF
--- a/exocom-server/CONTRIBUTING.md
+++ b/exocom-server/CONTRIBUTING.md
@@ -14,10 +14,11 @@ More details around configuring the port [here](features/configuring-the-ports.f
 ## Terminology
 
 - there are a number of __service types__, for example the "web" service, the "todo" service, etc
-- each service type has a number of instances. We call them __clients__ of Exocom to avoid the word "service" again
+- each service type has a number of instances. We call them __clients__ of Exocom to avoid the word "service" again.
+- each client has a unique name to identify it, e.g. "web #1"
 - the service configuration file defines which messages a service is allowed to send and receive.
   This information is called __routing__
-- services can send __messages__ to each other.
+- service instances (clients) can send __messages__ to each other.
   A service can only send messages of a __message type__ it is allowed to send,
   which is specified in the routing information.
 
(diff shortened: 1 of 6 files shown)


exocom-server/src/client-registry/client-registry.ls
```diff
@@ -38,19 +38,20 @@ class ClientRegistry
...
-  # Adds service routing configurations to the given setup
-  register-client: (service) ->
-    @clients[service.name] =
-      client-name: service.name
-      service-type: service.name
-      internal-namespace: @routing[service.name].internal-namespace
+  # registers the given service instance that just came online
+  register-client: (client) ->
+    @clients[client.name] =
+      client-name: client.name
```
trushton: Do we need a step here to make sure that the client name is available, and to alter it if it isn't? In #26 you mention that the client names could be essentially the `service-type` followed by an incremental number, e.g. `web1`, `web2`.

exocom-server/CONTRIBUTING.md
```diff
@@ -11,6 +11,18 @@ The default port is 3100.
 More details around configuring the port [here](features/configuring-the-ports.feature).
 
 
+## Terminology
+
+- there are a number of __service types__, for example the "web" service, the "todo" service, etc
+- each service type has a number of instances. We call them __clients__ of Exocom to avoid the word "service" again.
+- each client has a unique name to identify it, e.g. "web #1"
+- the service configuration file defines which messages a service is allowed to send and receive.
+  This information is called __routing__
```
trushton: In consistency with this, should the `messages` section of a service's `service.yml` be renamed to `routing`?

hugobho: I think a service stills sends individual `messages`, but `routing` is a list of what messages a service can send or receive.

kevgo: +1 on what @hugobho said. `service.yml` defines the messages that a service sends and receives. Routing is the table compiled out of that information that says which message should be forwarded to which roles.

exocom-server/src/client-registry/client-registry.ls
kevgo: Yes, absolutely! Let's do that once we build support for multiple instances of roles. That feature needs proper feature specs, tests, etc, so I don't want to do that here in this PR.

PR #44, opened 2017-01-06
KG 3: streamline subsystem architecture

kevgo:
This PR creates a proper subsystem architecture for Exocom:
- the kernel (exocom.ls) creates all subsystems
- subsystems don't know about the kernel nor other subsystems. They only know about themselves. They are given config information by the kernel, and emit events when something happens.
- the kernel listens to the events emitted by the subsystems. To handle them, the kernel tells the different subsystems to do things.
- no more circular dependencies between kernel and subsystems

@trushton


exocom-server/src/exocom.ls
```diff
@@ -34,6 +37,7 @@ class ExoCom extends EventEmitter
   # deregisters a service instance that went offline
   deregister-client: (client-name) ~>
     @client-registry.deregister-client client-name
+    @websocket.deregister-client client-name
```
kevgo: the kernel now contains the handler methods for all types of events. They orchestrate the different subsystems to handle the respective event.

exocom-server/src/exocom.ls
```diff
@@ -19,11 +19,14 @@ class ExoCom extends EventEmitter
...
     @http-subsystem = new HttpSubsystem {exocom: @, @logger}
       ..on 'online', (port) ~> @emit 'http-online', port
+      ..on 'config-request', (http-response) ~> @on-config-request http-response
 
     @message-cache = new MessageCache!
 
     @websocket = new WebSocketSubsystem {exocom: @, @logger}
       ..on 'online', (port) ~> @emit 'websockets-online', port
+      ..on 'deregister-client', (client-name) ~> @deregister-client client-name
+      ..on 'message', (message) ~> @handle-incoming-message message
```
kevgo: the kernel listens to the events emitted by the subsystems

PR #46, opened 2017-01-06
Explain domain model in main monorepo documentation

kevgo:
Before this was in the exocom-server subproject, but it applies to all subprojects here, so I'm moving it to the documentation for the main repo. Also fleshing out more details, please give this a thorough review and makes sure it conforms to what we discussed on the standup.

implements #35

@trushton @hugobho


CONTRIBUTING.md
```diff
@@ -12,6 +11,57 @@ This monorepository contains the different subprojects for it:
...
+To run the application, Exosphere creates an instance of the message bus,
+plus an instance of the respective service for each role.
+Or several instances for load balancing.
+These instances are called __clients__ (because they are clients of the message bus).
+Clients have unique names to identify them, for example "web #1", "web #2", etc.
+Clients communicate via the central __message bus__,
+played by the [Exocom-Server](exocom-server).
+
+The __application configuration__ file (`application.yml`) defines
+which roles the application has, and which _service type_ plays which role
```
hugobho: My understanding was that "service type" is defined in `service.yml`:
```
# users service service.yml

type: mongo # this is the service type
description: Manages user accounts in Exosphere applications
author: originate
```
and only the service role is defined in application.yml:
```
#application.yml

users: #service role
    location: ./exosphere-users-service
tweets: #service role
    location: ./exosphere-tweets-service
```

CONTRIBUTING.md
```diff
@@ -12,6 +11,57 @@ This monorepository contains the different subprojects for it:
...
+in which environment.
+
+Each service type contains a __service configuration__ file (`service.yml`),
```
hugobho: Can't the same service type have different looking `service.yml` files? For instance, both users and tweets are a `mongo` service type but send and receive different messages, and maybe have other differences in their service config.

CONTRIBUTING.md
```diff
@@ -12,6 +11,57 @@ This monorepository contains the different subprojects for it:
...
+which defines the __message types__ this service type will send and receive.
```
hugobho: Ah, message payload will be defined in `service.yml` as well now? Cool!

CONTRIBUTING.md
```diff
@@ -12,6 +11,57 @@ This monorepository contains the different subprojects for it:
...
+A _message type_ specifies the title, payload structure,
+and semantic meaning for message of that type.
+
+Both types of information combined create the __routing__ information for the application.
```
hugobho: It's unclear what is being referred to here by "both types", though it seems like it's clarified in the sentence after. Perhaps something along the lines of: 

Information from the **application configuration** and the **service configuration** combine to create the **__routing__** information for the application.

CONTRIBUTING.md
```diff
@@ -12,6 +11,57 @@ This monorepository contains the different subprojects for it:
...
+a _rest_ service for exposing it via a REST API,
+and a _chat_ service for exposing it via a chatbot.
+
+"User" and "todo" are the __role__ that the respective services play in the application.
+A role can be played by different __service types__.
+For example, you could use the _FacebookUserService_ service type
+to play the role of the _user_ service for your application in production,
+and you could use a _MockUserService_ to play the same role for local testing.
+You can also use the same service type for different roles,
+for example use a generic _MongoService_ as a quick-and-dirty prototype
```
trushton: The wording on this line seems weird, maybe something like this could work? 
`for example, *you could* use a generic...`

CONTRIBUTING.md
kevgo: Excellent catch! I reworded the whole section, please re-review.

CONTRIBUTING.md
kevgo: The service type here is `MongoService`. It sends and receives "mongo.*" messages, independent of whether it plays the role of "user" or "tweets" service in the app. That's why Exocom has to translate from "user.create" to "mongo.create".

CONTRIBUTING.md
kevgo: Yes, once we get to [milestone 0.4](https://github.com/Originate/exosphere-sdk/milestone/17)!

CONTRIBUTING.md
kevgo: Good idea. Reworded the section

CONTRIBUTING.md
kevgo: I like that. Changed.

CONTRIBUTING.md
hugobho: 👍

CONTRIBUTING.md
hugobho: It's still unclear to me what you mean by the second half of this sentence:
> and which _service type_ plays which role in which environment.

Can you give an example of what this looks like in `application.yml`? I thought the only thing `application.yml` knew about services was their service role and location. 

And what are you referring to when you say environment?

hugobho: Oh, by environment do you mean public or private?

kevgo: _Environment_ is `production`, `development`, etc. I have added wording to clarify that, this was missing.

PR #49, opened 2017-01-10
Build exoservice using Webpack

trushton:
`exorelay-javascript` is also pulled from the monorepo into this build, so `exoservice` does not have to depend on the published version of `exorelay`

The step of building with webpack was added to `bin/setup`

@kevgo @hugobho


exoservice-javascript/bin/exo-js
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env node
 
-require('../dist/cli');
+require('../dist/exo-js');
```
kevgo: Why does this change?

exoservice-javascript/bin/setup
```diff
@@ -1,3 +1,5 @@
 #!/usr/bin/env bash
 
 yarn install
+./node_modules/.bin/build
+webpack
```
kevgo: I thought webpack contains a livescript compiler. So there shouldn't be a separate transpilation step via `node_modules/bin/.build` here.

trushton: Thats true! I'll go ahead and remove that

exoservice-javascript/bin/exo-js
trushton: This changes so that we are running the output of the webpack build. Here I named it `exo-js` and placed it in the `dist` folder, but it could be named anything we want it to be.

kevgo: okay

kevgo:
Please also update the [developer documentation](https://github.com/Originate/exocom/blob/master/exoservice-javascript/CONTRIBUTING.md)

trushton:
@kevgo

exoservice-javascript/bin/setup
```diff
@@ -1,3 +1,4 @@
 #!/usr/bin/env bash
 
 yarn install
+webpack
```
kevgo: Please create an executable file `bin/build` that runs `webpack`. Then call `build` here. This way a developer can simply run `build` on the command line and it always does the right thing in each repo.

PR #51, opened 2017-01-12
Extract message translator class

hugobho:
resolves https://github.com/Originate/exosphere-sdk/issues/192

@kevgo @trushton


exocom-server/src/message-translator/message-translator.ls
```diff
@@ -0,0 +1,29 @@
...
+
+  # Returns the message name to which the given service would have to subscribe
+  # if it wanted to receive the given message expressed in its internal form.
+  #
+  # Example:
+  # - service "tweets" has internal namespace "text-snippets"
+  # - it only knows the "text-snippets.create" message
+  # - the external message name that it has to subscribe to is "tweets.create"
+  public-message-name: ({internal-message-name, client-name, internal-namespace}) ->
+    message-parts = internal-message-name.split '.'
```
kevgo: You could inline this variable to improve readability (allows you to get rid of the `switch` statement and do pure pattern matching in this method

exocom-server/src/message-translator/message-translator.ls
```diff
@@ -0,0 +1,29 @@
...
+    switch
+    | !internal-namespace              =>  internal-message-name
+    | message-parts.length is 1        =>  internal-message-name
+    | message-parts[0] is client-name  =>  internal-message-name
+    | otherwise                        =>  "#{client-name}.#{message-parts[1]}"
+
+
+  # Translates outgoing message into one that the receiving service will understand
+  internal-message-name: (message-name, {for: service}) ->
+    message-parts = message-name.split '.'
```
kevgo: inline this as well

hugobho:
Seems like some tests are red. At least for this PR, `exocom-server` tests pass for Node 7 but not 6. Has this been happening? Looks like it's related to an incorrect response time on an ExoCom reply message.

kevgo:
@trushton is still in the process of getting our tests green. I did some breaking changes on Exocom. Strange that it passes on Node 7 but not 6, though!

PR #56, opened 2017-01-21
ExoCom should use service type in routing config

hugobho:
SERVICE_ROUTES looks like this:
```
[{service-type: users-service, sends: [], receives:[], namespace: mongo}]
```
ExoCom was previously using client name (same as service role right now).

@kevgo


exocom-server/features/broadcasting-messages.feature
```diff
@@ -13,12 +13,12 @@ Feature: Broadcasting messages
     """
     [
       {
-        "role": "web",
+        "serviceType"": "web",
```
kevgo: double quotation here

exocom-server/features/support/mock-service.ls
```diff
@@ -19,7 +19,10 @@ class MockService
 
 
   connect: ({payload}, done) ~>
-    payload ?= {@client-name}
+    payload ?= {
```
kevgo: I think there is enough space to put this all on one line here

PR #59, opened 2017-01-23
Revert to 37d13a6

hugobho:
@kevgo 

We have been mistakenly changing service role to service type in ExoCom's routing configuration. This PR reverts those commits.


hugobho:
This PR is not reverting to the correct commit. Opening another PR.

PR #60, opened 2017-01-23
Revert to commit 0ff9719

hugobho:
@kevgo 

This PR reverts all commits related to changing service role to service type for ExoCom's SERVICE_ROUTING


exorelay-javascript/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "exorelay",
-  "version": "0.20.2",
+  "version": "0.20.1",
```
kevgo: leave the version at 2

exocom-server/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "exocom-dev",
-  "version": "0.16.2",
+  "version": "0.16.0",
```
kevgo: leave the version as is

exoservice-javascript/features/example-services/dependencies-contain-server-js/service.yml
```diff
@@ -1,3 +1,3 @@
-type: dependency-contains-server
+name: dependency-contains-server
```
kevgo: Isn't `type` more correct here?

hugobho: Yes, these should all be `type`

PR #62, opened 2017-01-24
Always test all repos on CI

kevgo:
As discussed today, let's test all repos at all times until we can specify dependencies more correctly.

@hugobho


hugobho:
We should do this on the exosphere monorepo as well, no?

PR #64, opened 2017-01-24
Include Exorelay packages in Exoservice

hugobho:
@kevgo @trushton 

Forgot to include these changes in the reversion we did. Fixes issue where services using Exoservice could not find the `uuid` module


exoservice-javascript/package.json
```diff
@@ -1,18 +1,22 @@
 {
   "name": "exoservice",
-  "version": "0.18.6",
+  "version": "0.18.3",
```
hugobho: I see this...
Making a PR to fix this. :man_facepalming:

PR #66, opened 2017-01-24
Fix webpack build for ci tests

trushton:
`exorelay` needed to be setup and built so that `exoservice` can access it correctly for building the webpack file.

@kevgo @hugobho


.travis.yml
```diff
@@ -17,6 +17,7 @@ before_install:
   - git checkout $TRAVIS_BRANCH
   - curl --output morula --location --fail https://github.com/Originate/morula/releases/download/0.1/morula-linux-amd64 && chmod +x morula
   - npm install -g yarn
+  - (cd exorelay-javascript && bin/setup && node_modules/.bin/build)
```
kevgo: We run `setup` and `build` in the `install` block on line 24. Why is this done here?

trushton: The exorelay `setup` script does not run build. Also, the main reason we need to do this here is that if changes were made to only exoservice, we still need to build exorelay on the CI server so that it is available for the webpack build. (similar to exosphere-shared always being built for exosphere).

Should exorelay be added to the `always` section of `morula.yml` and the build step added to its `bin/setup` script?

kevgo: Either `exo setup` or `exo run`/`exo test` should run build. There should be no need to run it manually.

Sounds like you need to add `exorelay` to the `before` block of `morula.yml`?

PR #69, opened 2017-01-25
Address comments from #66

trushton:
resolves #66 

This PR addresses the comments from the previously closed PR #66 

@kevgo @hugobho


exoservice-javascript/bin/spec_ci
```diff
@@ -2,7 +2,7 @@
 set -e
 
 node_modules/o-tools-livescript/bin/build
-webpack
+bin/build
```
trushton: The `bin/build` script in `exoservice-javascript` runs the `webpack` command.

morula.yml
```diff
@@ -1,4 +1,4 @@
-before-all: ""
+before-all: "exorelay-javascript"
```
trushton: This is required to be built so that `exoservice-javascript` has access to it for building the webpack bundle.

exoservice-javascript/bin/spec_ci
kevgo: Yes, let's use the tools we wrote to keep this DRY.

PR #71, opened 2017-01-26
Run build in spec

hugobho:
@kevgo @trushton


exoservice-javascript/bin/spec
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
+webpack
```
kevgo: Let's follow our convertion for helper scripts:
- `build` compiles source code into executable formats, for example via `lsc` or `webpack`
- `spec` calls `build`

hugobho:
Ready for another review. I've also made changes to the other subprojects to run `build` in `spec`

exocom-mock-javascript/bin/spec
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
-node_modules/o-tools-livescript/bin/build
+build
```
kevgo: I think this should be `bin/build`, since it will be called from the `exocom-mock-javascript` directory.

hugobho:
@kevgo comments addressed.

PR #81, opened 2017-02-04
Ensure in root directory before publish

hugobho:
ExoCom equivalent of this PR: https://github.com/Originate/exosphere/pull/210

@kevgo


bin/package.json
```diff
@@ -7,6 +7,7 @@
     "inquirer": "3.0.1",
     "livescript": "1.5.0",
     "o-tools-livescript": "1.2.3",
+    "path": "0.12.7",
```
kevgo: Use built-in path module

PR #82, opened 2017-05-01
Fix Exorelay tests

kevgo:
resolves parts of #250

A number of issues were messed up


kevgo:
Shipping this for now, please review anyways!

kevgo:
@hugobho it tried to send a reply after the websocket was closed. I think that came from some of these tests not using the right data, and therefore not properly waiting. Let's see.

PR #84, opened 2017-05-01
Update dependencies

kevgo:


exorelay-javascript/package.json
```diff
@@ -4,27 +4,27 @@
...
   },
   "description": "Communication relay between JavaScript code bases and the Exosphere environment",
   "devDependencies": {
     "chai": "3.5.0",
     "cucumber": "1.3.1",
     "cucumber-snippets-livescript": "1.0.1",
     "dependency-lint": "4.3.1",
-    "ejs": "2.5.5",
+    "ejs": "2.5.6",
     "exocom-mock": "0.21.4",
```
hugobho: Unrelated to this PR, but I think this makes versioning a bit tricky. If we make changes in both `exorelay` and `exocom-mock`, those changes will get published in `0.21.5`, but `exorelay@0.21.5` will have been testing against an older version of `exocom-mock@0.21.4`. 

This is a similar issue we had with `exo-relay` and `exo-service` both needing to be lock-versioned but also depending on each other's latest changes. Although maybe it's not as big of an issue since `exocom-mock` is only used for tests in development.

kevgo: Good point. I have created https://github.com/Originate/exosphere-sdk/issues/255 to track this!

PR #90, opened 2017-05-04
Run Go tests on Travis

kevgo:


kevgo:
@hugobho @trushton FYI, GOod times are ahead!

PR #91, opened 2017-05-04
Fix paths to Node programs

kevgo:
Small refactoring to prepare the code base to install Node manually as part of running Golang tests on Travis.


exorelay-javascript/bin/spec_ci
```diff
@@ -3,5 +3,5 @@ set -e
 
 node_modules/o-tools-livescript/bin/build
 bin/lint
-mocha --compilers ls:livescript "**/*spec.ls"
+node_modules/.bin/mocha --compilers ls:livescript "**/*spec.ls"
```
kevgo: @hugobho here are the changes for Exorela

PR #96, opened 2017-05-05
Initial working spike

alexdavid:
Sorry code isn't great


go/exorelay/exorelay_test.go
```diff
@@ -29,12 +32,18 @@ func FeatureContext(s *godog.Suite) {
...
 	s.Step(`^it registers by sending the message "([^"]*)" with payload:$`, func(expectedName string) error {
-		messages := exocom.ReceivedMessages
+		var messages []structs.Message
+		for {
+			time.Sleep(time.Duration(10) * time.Millisecond)
+			messages = exocom.ReceivedMessages
+			if len(messages) == 1 {
+				break
+			}
+		}
```
alexdavid: Maybe this functionality can go into exocom mock? Maybe a `waitForReceivedMessage` that returns a message, error and times out after a period and returns an error

go/exorelay/main.go
```diff
@@ -1,10 +1,49 @@
...
+	socket, err := websocket.Dial(exoRelay.url, "", "origin:")
+	if err != nil {
+		return err
+	}
+	exoRelay.socket = socket
+	err = exoRelay.Send("exocom.register-service", nil)
+	if err != nil {
+		return err
+	}
+	return nil
```
alexdavid: I'm not sure which is more idiomatic, but we should stick to one style. Later on we return the error directly so this should either be
```go
return exoRelay.Send("exocom.register-service", nil)
```

or we should add the error handling shenanigans below

go/exorelay/exorelay_test.go
kevgo: Yes, good idea! That's how the JS implementation works. For now we can do without timeout, since the test framework times out (hopefully).

go/exorelay/main.go
kevgo: Yes, we can simplify the code to what you propose. It does the same thing.

trushton:
@alexdavid @kevgo I addressed the two comments above. Is this PR to be merged before expanding out to include more tests? Or should they be included here?

PR #97, opened 2017-05-07
Remove dev dependencies from vendor

kevgo:
@alexdavid I read that libraries should not vendor dependencies:

http://glide.readthedocs.io/en/latest/vendor:
> Libraries (codebases without a main package) should not store outside packages in a vendor/ folder in their VCS unless they have a specific reason and understand why they're doing it.

I am afraid this will break development when there is a new version of godog, and I don't have a good answer for this, but it's probably better to follow best practices here.


alexdavid:
This is interesting. I'd imagine dev dependencies could still be vendored, no? Maybe we can use a submodule for it?

kevgo:
I don't have enough experience, just following what is written out in https://peter.bourgon.org/go-best-practices-2016/#dependency-management

kevgo:
I can close this PR if you don' want to do this right now, @alexdavid. Not feeling strongly about this.

kevgo:
Does it include the vendored dependencies into the compiled binary?

PR #98, opened 2017-05-16
go exorelay: test register-service payload

charlierudolph:
The payload was not being tested against.


go/exorelay/exorelay_test.go
```diff
@@ -19,13 +20,12 @@ func FeatureContext(s *godog.Suite) {
...
 
-	s.Step(`^an ExoRelay with the configuration:$`, func(configStr *gherkin.DocString) error {
-		var config map[string]interface{}
-		err := json.Unmarshal([]byte(configStr.Content), &config)
-		exoInstance = New("ws://localhost:4100", config)
-		if err != nil {
-			return err
-		}
+	s.Step(`^an ExoRelay with the role "([^"]*)"$`, func(role string) error {
+		exoInstance = New(ExoRelayConfig{
```
kevgo: What do you think about renaming this package to "exocom_test", so that we have to test the public interface of Exocom?

charlierudolph: I don't understand your comment. This is in the exorelay folder. I'm good adding tests for the exocom-mock (which I will expect we will need anyway and the javascript version has).

kevgo: Sorry I meant `exorelay_test`. I have created https://github.com/Originate/exocom/pull/103 to address this comment, so please go ahead and ship this as is. Thanks!

PR #101, opened 2017-05-16
Exocom: reduce polling interval time

kevgo:
Waiting 1ms is enough, makes the tests a bit snappier


go/exocom-mock/exocom-mock.go
```diff
@@ -61,7 +61,7 @@ func (exoCom *ExoComMock) WaitForReceivedMessage() (structs.Message, error) {
 
 	go func() {
 		for {
-			time.Sleep(time.Duration(10) * time.Millisecond)
+			time.Sleep(time.Duration(1) * time.Millisecond)
```
charlierudolph: can this be simplified to
```go
time.Sleep(time.Millisecond)
```

PR #102, opened 2017-05-16
Standardized tool setup for Go code

kevgo:
- sets up the standardized [O-Tools](https://github.com/originate/o-tools-node) toolset
- uses gometalinter
- fixes the lint errors


go/exorelay/exorelay_test.go
```diff
@@ -17,16 +17,18 @@ import (
 func FeatureContext(s *godog.Suite) {
 	var exoInstance *ExoRelay
 	exocom := exocomMock.New()
-	go exocom.Listen(4100)
+	go func() {
+		err := exocom.Listen(4100)
+		if err != nil {
+			fmt.Printf("ERROR: cannot start exocomMock: %s", err)
```
kevgo: This is my quick and dirty attempt at checking the return value of `exocom.Listen`, which the linter correctly calls us out for. Strange that the type checker didn't complain about this.

go/exocom-mock/bin/spec
```diff
@@ -1,7 +1,4 @@
 #!/usr/bin/env bash
 set -e
 
-echo "no tests so far"
-exit
 bin/lint
-go install && go test
```
trushton: don't we still want `go test` here?

go/exocom-mock/bin/setup
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
 set -e
 
-echo "no setup necessary"
+glide install
+go get go get -u gopkg.in/alecthomas/gometalinter.v1
```
trushton: remove one of the two `go get`s

go/exorelay/exorelay_test.go
charlierudolph: Thoughts on `log.Fatal` here? I don't think we want to continue running if this fails

PR #104, opened 2017-05-16
go exorelay: update dependencies and add 'go.uuid'

charlierudolph:


go/exorelay/glide.lock
```diff
@@ -1,16 +1,13 @@
...
-  version: 38e3d1fb6df6a7b6c59c120ecd02141764d2c343
+  version: 92fbee719c9593813b24a6ef042bcad40324ede4
   subpackages:
   - colors
   - gherkin
-- name: github.com/Originate/exocom
-  version: 30c7f513ced53c2817f0f41aac60322e7eeef069
-  subpackages:
-  - go/exorelay
-  - go/structs
```
charlierudolph: We don't want these to actually get installed. Glide will add them automatically when you get a new package. I've just manually deleted them.

go/exorelay/glide.yaml
```diff
@@ -1,4 +1,7 @@
 package: github.com/Originate/exocom/go/exorelay
 import:
 - package: github.com/DATA-DOG/godog
-  version: v0.7.0
+  version: v0.7.2
+- package: golang.org/x/net
```
charlierudolph: This got added the the glide.lock file automatically since it realized it was a dependency. Making it more explicit here

go/exorelay/glide.lock
kevgo: Argh, this fixes a long-standing headache where suddenly it installs a copy of exocom in vendor and then doesn't use the real version anymore. Thanks for fixing this!

PR #105, opened 2017-05-16
go exorelay: add sending tests

charlierudolph:


go/exocom-mock/exocom-mock.go
```diff
@@ -37,14 +38,18 @@ func (exoCom *ExoComMock) Listen(port int) error {
...
+func (exoCom *ExoComMock) websocketHandler(socket *websocket.Conn) {
+	for {
+		message, err := exoCom.readMessage(socket)
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			panic(err)
+		} else {
+			exoCom.ReceivedMessages = append(exoCom.ReceivedMessages, message)
+		}
```
charlierudolph: Previously we were only reading the first message. Took me a while to figure this out

go/exorelay/exorelay_test.go
```diff
@@ -16,26 +16,26 @@ import (
...
+	exocom := exocomMock.New()
+	go func() {
+		err := exocom.Listen(4100)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}()
+	defer exocom.Close()
+	return exocom
+}
```
charlierudolph: Couldn't end up using a new exocom per scenario as you get the error "Server closed". Went with this and reset the received messages every test.

go/exorelay/features/sending.feature
```diff
@@ -0,0 +1,49 @@
...
+          "name": "world"
+        },
+        "sender": "test-service",
+        "id": "{{request_uuid}}"
+      }
+      """
+
+  Scenario: trying to send an empty message
+    When trying to send an empty message
+    Then ExoRelay errors with "ExoRelay#Send cannot send empty messages"
```
charlierudolph: Far fewer features here then the JS tests since a lot of them were for type checking. The JS version does allow payload to be a string or an array. That seems harder with types. Is that a requirement?

go/exorelay/exorelay_test.go
```diff
@@ -16,26 +16,26 @@ import (
 	"github.com/Originate/exocom/go/structs"
 )
 
+func initializeExocom() *exocomMock.ExoComMock {
```
kevgo: Thoughts about going with the Go vocabulary and naming this "newExocom"?

go/exorelay/exorelay_test.go
```diff
@@ -16,26 +16,26 @@ import (
...
+	exocom := exocomMock.New()
+	go func() {
+		err := exocom.Listen(4100)
```
kevgo: Why is this in a goroutine? Is it blocking?

go/exorelay/features/sending.feature
kevgo: Good point. Let's go with any payload that is valid JSON. That would work with static typing as well.

go/exorelay/exorelay_test.go
kevgo: I have yet to see a platform that can tolerate opening and closing a lot of sockets on the same port. Too much heavyweight OS stuff underneath I guess.

go/exorelay/exorelay_test.go
charlierudolph: Yep. It was before as well

go/exocom-mock/exocom-mock.go
```diff
@@ -64,26 +69,28 @@ func (exoCom *ExoComMock) readMessage(socket *websocket.Conn) (structs.Message,
...
-// WaitForReceivedMessage blocks until this instance receives the given message
-func (exoCom *ExoComMock) WaitForReceivedMessage() (structs.Message, error) {
-	c1 := make(chan structs.Message, 1)
-	numMessages := len(exoCom.ReceivedMessages)
+func (exoCom *ExoComMock) Reset() {
+	exoCom.ReceivedMessages = []structs.Message{}
+}
+
+// WaitForReceivedMessage blocks until this instance receives the given number of messages
+func (exoCom *ExoComMock) WaitForReceivedMessage(count int) error {
```
charlierudolph: sounds good

go/exorelay/exorelay_test.go
charlierudolph: sounds good

go/exocom-mock/exocom-mock.go
charlierudolph: went with `WaitForReceivedMessageCount`

go/exorelay/exorelay_test.go
charlierudolph: fixed

charlierudolph:
@kevgo all comments addressed

go/exocom-mock/exocom-mock.go
```diff
@@ -37,14 +38,18 @@ func (exoCom *ExoComMock) Listen(port int) error {
...
-	if err != nil {
-		panic(err)
+// websocketHandler is called when the given socket connects
+func (exoCom *ExoComMock) websocketHandler(socket *websocket.Conn) {
+	for {
+		message, err := exoCom.readMessage(socket)
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			panic(err)
```
kevgo: Do we really want to crash the entire running test here? What about printing the error and just breaking the loop? I can see such errors happening when ramping down tests.

go/exorelay/exorelay_test.go
```diff
@@ -16,26 +16,26 @@ import (
...
 
+func newExocom() *exocomMock.ExoComMock {
+	exocom := exocomMock.New()
+	go func() {
+		err := exocom.Listen(4100)
+		if err != nil {
+			log.Fatal(err)
+		}
+	}()
+	defer exocom.Close()
```
kevgo: I think `defer` closes the exocomMock instance once this method returns. Is that what we want here?

go/exorelay/exorelay_test.go
```diff
@@ -71,6 +71,53 @@ func FeatureContext(s *godog.Suite) {
...
+		err := json.Unmarshal([]byte(payloadStr.Content), &payload)
+		if err != nil {
+			return err
+		}
+		return exoInstance.Send(message, payload)
+	})
+
+	s.Step(`^trying to send an empty message$`, func() error {
+		savedError = exoInstance.Send("", nil)
+		return nil
```
kevgo: In the current Cucumber paradigm, if the step says "When trying to xxx", it is expected that the activity xxx fails. If it succeeds, this step should fail. We can check the exact error in a subsequent step, but here we should check that an error has occurred at all. So, what do you think about adding this line:

```go
if err == nil {
  return fmt.errorf("Expected error, got none")
} else {
  return nil
}
```

I have run into a number of situations where it was expected that an error happens like this, the error didn't occur because of a bug in the code, but the step was still passing in Cucumber. That was always misleading and made debugging harder.

go/exorelay/features/sending.feature
```diff
@@ -0,0 +1,49 @@
...
+
+  As an Exoservice developer
+  I want my service to be able to send messages to other Exosphere services
+  So that it can interact with the rest of the application.
+
+  Rules:
+  - call "Send" on your ExoRelay instance have it send out the given message
+  - provide the message to send as the first parameter
+  - provide payload for the message as the second parameter
+  - the payload can be either a string, an array, or a Hash
```
kevgo: I think we changed this to "the payload is a JSON structure"

go/exorelay/features/sending.feature
```diff
@@ -0,0 +1,49 @@
+Feature: Sending outgoing messages
+
+  As an Exoservice developer
+  I want my service to be able to send messages to other Exosphere services
+  So that it can interact with the rest of the application.
+
+  Rules:
+  - call "Send" on your ExoRelay instance have it send out the given message
+  - provide the message to send as the first parameter
```
kevgo: Let's refine this to: "provide the __name__ of the message to send as the first parameter"

go/structs/message.go
```diff
@@ -2,6 +2,8 @@ package structs
 
 // Message defines the structure of websocket packets representing messages
 type Message struct {
-	Name    string
-	Payload map[string]interface{}
+	Id      string                 `json:"id"`
```
kevgo: Very good idea to use tags for JSON encoding!

go/exorelay/exorelay_test.go
charlierudolph: Nope. I'll move it to after suite.

go/exorelay/exorelay_test.go
charlierudolph: Sounds good

go/exocom-mock/exocom-mock.go
charlierudolph: sounds good

PR #108, opened 2017-05-17
fix CI builds to catch errors

charlierudolph:
`bin/spec_ci` was missing `set -e` so errors were able to pass through.


go/exorelay/bin/lint
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
 set -e
 
+go install
```
charlierudolph: One of the linters needs this. See here: https://github.com/golang/go/issues/12703#issuecomment-144429128

bin/spec_ci
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -ex
```
charlierudolph: `set -x` prints out the commands run which helps parse what output is from what

go/exorelay/bin/lint
kevgo: Can you please add this as a comment to this line in all files? Future generations of developers who don't see this PR might wonder about this as well  :)

PR #109, opened 2017-05-17
go exorelay: message handlers

charlierudolph:


go/exorelay/exorelay_test.go
```diff
@@ -126,6 +164,56 @@ func FeatureContext(s *godog.Suite) {
...
 		return nil
 	})
+
+	s.Step(`^I register a handler for the "([^"]*)" message$`, func(messageName string) error {
+		mapping := map[string]func(map[string]interface{}){}
+		mapping[messageName] = func(payload map[string]interface{}) {
+			handlerCalled = true
+			savedPayload = payload
+		}
+		exoInstance.BindMessageHandlers(mapping)
```
charlierudolph: Here is an example of binding message handlers. You do them all at once, supplying a map from the message name to a function. This is similar to how @alexdavid handled a similar problem on pairfoce.

go/exorelay/main.go
```diff
@@ -57,3 +66,42 @@ func (exoRelay *ExoRelay) Send(eventName string, payload map[string]interface{})
...
+	}
+
+	var unmarshaled structs.Message
+	err := json.Unmarshal(bytes, &unmarshaled)
+	if err != nil {
+		return structs.Message{}, err
+	}
+
+	return unmarshaled, nil
+}
```
charlierudolph: `readMessage` is an exact duplicate of what lives in exocom-mock. Because `websocket` is a dependency, I get an error about different types as the definition for the socket being passed is in separate places.

Thus to remove this duplication, I think we would need to encapsulate that entire library (something I attempted to do this morning and then threw out as I don't think its worth it. Could still refactor down the road.

charlierudolph: As a potential way to fix this, thoughts on bringing the `exocom-mock` (and `structs`) into the `exorelay` folder. Unless they need to be their own project, it would be easier to develop if they all were under `exorelay`.

go/exocom-mock/exocom-mock.go
```diff
@@ -16,6 +15,7 @@ import (
 type ExoComMock struct {
 	ReceivedMessages []structs.Message
 	server           http.Server
+	serviceSockets   map[string]*websocket.Conn
```
kevgo: Do we really need infrastructure to store several websockets? Since this is only a mock Exocom for testing one particular ExoRelay instance, I only ever expect one websocket to connect to it.

So, can this just be:

```go
socket *websocket.Conn
```

go/exorelay/exorelay_test.go
```diff
@@ -27,16 +29,51 @@ func newExocom() *exocomMock.ExoComMock {
 	return exocom
 }
 
+func WaitFor(condition func() bool, errorMessage string) error {
```
kevgo: This is a very cool method! I am surprised it does not exist as an open source tool or framework method already.

go/exorelay/exorelay_test.go
```diff
@@ -27,16 +29,51 @@ func newExocom() *exocomMock.ExoComMock {
...
+
+	select {
+	case <-success:
+		return nil
+	case <-time.After(time.Second * 10):
+		return errors.New(errorMessage)
+	}
+}
+
+func WaitForReceivedMessagesCount(exocom *exocomMock.ExoComMock, count int) error {
```
kevgo: Thoughts about making this a method of `ExocomMock`? Then it could be called like this:

```go
exocom.WaitForReceivedMessagesCount(2)
```

go/exorelay/exorelay_test.go
kevgo: I don't get this, you have to walk me through this in the office

go/exorelay/main.go
kevgo: This might be because the `websocket` in ExocomMock is a different version than the one used here. And why it is recommended that libraries don't vendor their dependencies, but use the one from the binary that uses them. https://github.com/Originate/guide_private/pull/61 proposes this rule, based on literature in the Go community. Does deleting the vendor folder in ExocomMock solve this issue?

go/exorelay/exorelay_test.go
charlierudolph: It was there originally. That would require `WaitFor` to be moved into another package to prevent a circular import

go/exocom-mock/exocom-mock.go
charlierudolph: It was done that way over on the javascript version. Based on the features, yes it appears the map is unneeded

go/exorelay/exorelay_test.go
charlierudolph: There might be, I didn't look for anything as it was basically implemented in on exocom-mock.

go/exorelay/main.go
charlierudolph: In order for exocomMock to use the same as version as exorelay, it would need to be in the vendor folder of exorelay or a child of exorelay.

go/exorelay/exorelay_test.go
kevgo: We could use https://github.com/matryer/try instead of rolling our own WaitFor helper.

go/exorelay/exorelay_test.go
kevgo: https://github.com/matryer/try

go/exorelay/exorelay_test.go
```diff
@@ -130,6 +167,56 @@ func FeatureContext(s *godog.Suite) {
 		}
 		return nil
 	})
+
+	s.Step(`^I register a handler for the "([^"]*)" message$`, func(messageName string) error {
+		mapping := map[string]func(map[string]interface{}){}
+		mapping[messageName] = func(payload map[string]interface{}) {
+			handlerCalled = true
```
kevgo: We distinguish handlers by name, but `handlerCalled` is used for all handlers.

go/exorelay/exorelay_test.go
kevgo: Looking at this again, what is the motivation for having to provide all message handlers all at once here? Maps are mutable, why can't we provide an API that allows registering one handler at a time?

```go
exoInstance.bindMessageHandler(messageName, func(payload map[string]interface{}) {
  handlerCalled = true
  savedPayload = payload
}

File: go/exorelay/features/receiving.feature [hunk 7]
```diff
@@ -0,0 +1,46 @@
...
+      """
+      {
+        "name": "hello",
+        "payload": {
+          "name": "world"
+        },
+        "id": "123"
+      }
+      """
+    Then ExoRelay runs the registered handler, passing the following as the first argument
```
kevgo: We have the convention that if a step has an associated text block or table, it has a trailing `:`.

```cucumber
Then ExoRelay runs the registered handler with the argument:

File: go/exorelay/main.go [hunk 8]
```diff
@@ -19,8 +21,9 @@ type Config struct {
 
 // ExoRelay is the Go API to talk to Exocom
 type ExoRelay struct {
-	config Config
-	socket *websocket.Conn
+	config                Config
+	messageHandlerMapping map[string]func(map[string]interface{})
```
kevgo: Thoughts about naming this `messageHandlers`?

go/exorelay/exorelay_test.go
```diff
@@ -130,6 +167,56 @@ func FeatureContext(s *godog.Suite) {
 		}
 		return nil
 	})
+
+	s.Step(`^I register a handler for the "([^"]*)" message$`, func(messageName string) error {
+		mapping := map[string]func(map[string]interface{}){}
```
kevgo: I think handlers should receive `structs.Message` as their argument, not `map[string]interface{}`.

kevgo: Correction, I see that in your next PR you define a MessagePayload. That's the right approach!

go/exorelay/exorelay_test.go
charlierudolph: @alexdavid showed me an example and we couldn't see any reason why the user wouldn't supply them all at once. 

If we are specifying one by one, then we either need error checking that you can only specify one handler per message or need to implement an event emitter.

go/exorelay/exorelay_test.go
charlierudolph: handlerCalled is not called for all handlers. It is set for this one specific handler though. We could make it a map, but I don't think that would be worth it.

go/exorelay/features/receiving.feature
charlierudolph: updated

go/exorelay/main.go
charlierudolph: done

go/exorelay/exorelay_test.go
charlierudolph: I don't like the look of that library. It has retry limit and has you introduce your own delays. Our implementation has a timeout with a delay built in.

go/exorelay/exorelay_test.go
kevgo: I'd go with error checking. I think the resulting API is a lot more developer friendly. Similar to how we use query builders to build up massive SQL string constructs piece by piece.

go/exorelay/main.go
kevgo: The problem is that we will create an exoservice code base in the furure, and it will also need an exocommock and exorelay. 

If we can make it so that the exoservice code base can contain everything, and export a CLI tool and the ExoRelay library and the ExocomMock library, that would solve our problem. 

Meaning people must be able to do this:

Import "github.com/Originate/exocom/go/exoservice/exorelay" and get the ExoRelay. 

Import "github.com/Originate/exocom/go/exoservice/exocommock" and get the exocommock library.

kevgo: More background: https://forum.golangbridge.org/t/vendoring-and-multiple-versions-of-the-same-library/2010

go/exorelay/exorelay_test.go
kevgo: Okay, let's extract this into our own library then (in a separate PR).

go/exocom-mock/exocom-mock.go
kevgo: Let's keep it simple for now?

charlierudolph:
Closing this in favor of building a lower level API using channels. Will keep this branch around to use it in exoservice.

PR #110, opened 2017-05-17
go exorelay: outgoing replies

charlierudolph:


go/exorelay/features/outgoing-replies.feature
```diff
@@ -0,0 +1,44 @@
...
+    to send a reply to the message you are currently processing
+
+
+  Background:
+    Given an ExoRelay with the role "test-service"
+    And ExoRelay connects to Exocom
+
+
+  Scenario: sending a reply
+    Given I register a handler for "users.create" which replies with "users.created" and the payload:
```
kevgo: This example is lacking the relevant parts, i.e. how the API is to send a reply looks like. They are buried into the step definition. It would be nice to see this more directly somehow. In a compiled language that's not really possible, unfortunately.

We could create an `examples/send-reply.go` file that defines the handler. Then load it here:

```cucumber
Given I register the "examples/send-reply" handler for "users.create" 
When receiving this message: ...
```

This way, it is at least possible to see how the handler looks like.

go/structs/message.go
```diff
@@ -2,8 +2,12 @@ package structs
...
 type Message struct {
-	ID      string                 `json:"id"`
-	Name    string                 `json:"name"`
-	Payload map[string]interface{} `json:"payload"`
-	Sender  string                 `json:"sender"`
+	ID         string         `json:"id"`
+	Name       string         `json:"name"`
+	Payload    MessagePayload `json:"payload"`
+	Sender     string         `json:"sender"`
+	ResponseTo string         `json:"response-to"`
```
alexdavid: I think this json key is supposed to be camel case. Livescript changes the kabob case into camel case

go/exorelay/main.go
```diff
@@ -19,10 +19,16 @@ type Config struct {
 	Role string
 }
 
+// MessageHandler is the function signature for functions used for handling messages
+type MessageHandler func(structs.MessagePayload, func(string, structs.MessagePayload))
+
+// MessageHandlerMapping is a mapping from `message name to handler functions
+type MessageHandlerMapping map[string]MessageHandler
```
alexdavid: 👍 Way better

go/exorelay/main.go
```diff
@@ -19,10 +19,16 @@ type Config struct {
 	Role string
 }
 
+// MessageHandler is the function signature for functions used for handling messages
+type MessageHandler func(structs.MessagePayload, func(string, structs.MessagePayload))
```
alexdavid: Instead of giving a reply function, does a channel make more sense?

charlierudolph: Can you give a code example of the interface you are thinking? I've seen a couple use of channels but don't know how we would build an API with them

charlierudolph:
Closing this in favor of building a lower level API using channels. Will keep this branch around to use it in exoservice.

PR #112, opened 2017-05-18
reduce CI output

### LINKED ISSUE #106: reduce output on CI
Its terrible to look through. 

* dependeny-lint - upgrade to 5.0 where no it outputs nothing if there are no errors.
* cucumber - use progress formatter
* remove debug data that gets logged 
* Use `set -x` to be able to see what commands output what.

charlierudolph:
Old: https://travis-ci.org/Originate/exocom/builds/233442156
New: https://travis-ci.org/Originate/exocom/builds/233492950

resolves #106


exocom-server/bin/features
```diff
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
+set -ex
 
 node_modules/o-tools-livescript/bin/build
-if [ "$#" == "0" ]; then
-  node_modules/.bin/cucumber-js --tags ~@todo
+if [ "$#" != "1" ]; then
```
kevgo: The original intent was that when running `features`, it prints no output. When running `features foo.feature`, it prints debug output. I'd like to keep this behavior, since it is helpful for debugging. If we need something else on CI, let's modify `spec_ci`, it defines the exact commands to run on CI.

exocom-server/bin/spec_ci
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
-set -e
+set -ex
 
 node_modules/o-tools-livescript/bin/build
 node_modules/o-tools/bin/lint
-node_modules/.bin/mocha --compilers ls:livescript "**/*spec.ls"
-bin/features --format pretty
+node_modules/.bin/mocha --compilers ls:livescript "**/*spec.ls" --reporter dot
+bin/features --format progress
```
kevgo: Let's change this to `node_modules/.bin/cucumber-js --tags ~@todo --format progress` to avoid messing up the toolchain for development.

exocom-server/features/support/world.ls
```diff
@@ -21,7 +21,11 @@ World = !->
       PORT : port
       SERVICE_ROUTES : service-routes
     @port = port
-    @process = new ObservableProcess "bin/exocom", stdout: dim-console.process.stdout, stderr: dim-console.process.stderr, env: env
+    @process = new ObservableProcess "bin/exocom", {
+      stdout: if process.env.CI then false else dim-console.process.stdout
```
kevgo: Are you sure this doesn't remove important debug output?

go/exorelay/exorelay_test.go
```diff
@@ -142,7 +142,7 @@ func TestMain(m *testing.M) {
 	status := godog.RunWithOptions("godogs", func(s *godog.Suite) {
 		FeatureContext(s)
 	}, godog.Options{
-		Format:        "pretty",
+		Format:        "progress",
```
kevgo: Can we at least make it so that it uses `progress` when running all tests, but uses `pretty` when running a single test? The check for that condition is on line 137.

javascript/exoservice/package.json
```diff
@@ -23,7 +23,7 @@
     "chai": "3.5.0",
     "cucumber": "1.3.1",
     "cucumber-snippets-livescript": "1.0.1",
-    "dependency-lint": "4.3.1",
+    "dependency-lint": "^5.0.1",
```
kevgo: Please don't use floating versions.

exocom-server/bin/build
```diff
@@ -1,3 +1,4 @@
 #!/usr/bin/env bash
+set -ex
```
kevgo: Do we really need to print the command when running `build`? This command runs only one other command. It should be pretty obvious what command is going wrong in case there is an error.

exocom-server/bin/spec
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -e
+set -ex
```
kevgo: This isn't run on CI, only on developer machines. I'd like to keep the noise that this change introduces away from developers. There was never an issue with a developer not knowing what exact command failed.

exocom-server/bin/tests
```diff
@@ -1,3 +1,4 @@
 #!/usr/bin/env bash
+set -ex
```
kevgo: This runs only one command, I think this change is unnecessary here.

go/exocom-mock/bin/lint
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -e
+set -ex
```
kevgo: I think this particular change is unnecessary, and should only be added if there is a good reason. I run this command a lot on my dev machine, and don't want to be bothered with all this logging output.

go/exocom-mock/bin/spec
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
-set -e
+set -ex
```
kevgo: This does not run on CI, let's not change it.

go/exorelay/bin/lint
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -e
+set -ex
```
kevgo: Does not run on CI

go/exorelay/bin/spec
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -e
+set -ex
```
kevgo: does not run on CI

javascript/exorelay/bin/build
```diff
@@ -1,3 +1,4 @@
 #!/usr/bin/env bash
+set -ex
```
kevgo: This seems unnecessary

javascript/exorelay/bin/lint
```diff
@@ -1,3 +1,4 @@
 #!/usr/bin/env bash
+set -ex
```
kevgo: unnecessary

javascript/exorelay/bin/spec
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -e
+set -ex
```
kevgo: not run on CI

javascript/exorelay/bin/tests
```diff
@@ -1,3 +1,4 @@
 #!/usr/bin/env bash
+set -ex
```
kevgo: unnecessary

javascript/exoservice/bin/build
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -ex
```
kevgo: unnecessary

javascript/exoservice/bin/cuc-api
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -ex
```
kevgo: I'm not sure this is adding that much value either. Do you really want to see debug output for all these `if` clauses? All it does is call cucumber-js.

javascript/exoservice/bin/cuc-cli
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -ex
```
kevgo: same here

charlierudolph:
I'm going to split this into multiple PRs as I am making 4 distinct changes. Also in the future, if you want to comment on a change I made in multiple places, you can just comment in one place and note that is applies elsewhere instead of commenting on every instance.

PR #113, opened 2017-05-18
upgrade dev dependency: dependency-lint

charlierudolph:
Extracted from #112, reduces the output on CI as dependency-lint now by default outputs nothing if there are no errors.

@kevgo one of your comments
> Please don't use floating versions.

I switched to a locked down version but why does this matter if we are using yarn? Mainly the default version yarn installs has the carat and this is just extra work to remove it and rerun yarn install.


kevgo:
Floating versions are an antipattern in my opinion, and the source of instability. That's especially true in the presence of a lock file. They are confusing at best, and prevent updates to the latest version at worst. You can run `yarn config set save-prefix false` to prevent it from creating floating versions.

charlierudolph:
Thanks for sharing the config option!

PR #117, opened 2017-05-19
go exorelay: receive messages with channels

charlierudolph:


go/exocom-mock/exocom-mock.go
```diff
@@ -38,23 +38,31 @@ func (exoCom *ExoComMock) Listen(port int) error {
 	return exoCom.server.ListenAndServe()
 }
 
-// websocketHandler is called when the given socket connects
```
charlierudolph: This function just moved to the bottom of the file. Keeping all the public functions up top and putting the private ones at the end

go/exorelay/examples/index.go
```diff
@@ -0,0 +1,23 @@
...
+
+type Example interface {
+	GetReceivedMessages() []structs.Message
+	Setup(exoRelay *exorelay.ExoRelay)
+}
+
+func Get(name string) Example {
+	switch name {
+	case "message-handler":
+		return &MessageHandlerExample{}
```
alexdavid: What other types of examples do you foresee being here?

If you do want to keep this I suggest renaming this directory to `test_fixutres` or something because I was thinking these are code examples of how to use exorelay and this seems a bit over-engineered for that

go/exocom-mock/exocom-mock.go
```diff
@@ -38,23 +38,31 @@ func (exoCom *ExoComMock) Listen(port int) error {
...
-		if err == io.EOF {
-			break
-		} else if err != nil {
-			fmt.Println("Error reading message from websocket:", err)
-			break
-		} else {
-			exoCom.ReceivedMessages = append(exoCom.ReceivedMessages, message)
-		}
+// HasConnection returns whether or not a socket is connected
+func (exoCom *ExoComMock) HasConnection() bool {
```
charlierudolph: sounds good. I like it

go/exorelay/examples/index.go
charlierudolph: These are test fixtures used just for feature tests. I'll rename it.

The other two features I had envisioned implementing were [incoming-replies.feature](https://github.com/Originate/exocom/blob/7eec12403b1a18d02037bb680ab00556fbb7b65f/javascript/exorelay/features/incoming-replies.feature) and [outgoing-replies.feature](https://github.com/Originate/exocom/blob/7eec12403b1a18d02037bb680ab00556fbb7b65f/javascript/exorelay/features/outgoing-replies.feature). How do you think those should be dealt with? I'm not certain what to do with the outgoing reply aside from exposing a helper for the user to check if a message is a reply to another

charlierudolph: I'd also like to move the sending feature into an example to make it easier to see

go/exorelay/features/receiving.feature
```diff
@@ -0,0 +1,44 @@
+Feature: Receiving messages
+
+  As an Exosphere developer
+  I want my code base to be able to respond to incoming messages
+  So that I can create Exoservices.
+
+  Rules:
+  - you can listen for messages on a channel
```
kevgo: Incoming messages are provided via a channel

PR #118, opened 2017-05-20
go: cleanup

charlierudolph:
from comments on #117


go/exocom-mock/exocom-mock.go
```diff
@@ -28,42 +30,62 @@ func New() *ExoComMock {
 }
 
 // Close takes this ExoComMock instance offline
-func (exoCom *ExoComMock) Close() error {
-	return exoCom.server.Close()
+func (e *ExoComMock) Close() error {
+	return e.server.Close()
```
alexdavid: 👍

go/exocom-mock/exocom-mock.go
```diff
@@ -28,42 +30,62 @@ func New() *ExoComMock {
...
 
-// Reset nils the socket and clears all received messages
-func (exoCom *ExoComMock) Reset() {
-	exoCom.socket = nil
-	exoCom.ReceivedMessages = []structs.Message{}
+// Reset closes and nils the socket and clears all received messages
+func (e *ExoComMock) Reset() {
+	if e.HasConnection() {
+		err := e.socket.Close()
+		if err != nil {
```
alexdavid: What do you think about joining these lines?

```go
if err := e.socket.Close(); err != nil {
```

go/exorelay/main.go
```diff
@@ -35,25 +35,25 @@ func New(config Config) *ExoRelay {
 }
 
 // Connect brings an ExoRelay instance online
-func (exoRelay *ExoRelay) Connect() error {
-	socket, err := websocket.Dial("ws://"+exoRelay.config.Host+":"+exoRelay.config.Port, "", "origin:")
+func (e *ExoRelay) Connect() error {
+	socket, err := websocket.Dial("ws://"+e.config.Host+":"+e.config.Port, "", "origin:")
```
alexdavid: Is there a reason why `e.config.Port` is a string here?

go/exocom-mock/exocom-mock.go
charlierudolph: I like that idea when the function call is short but we have many other similar instances where the function call is long. For consistency, I'd prefer to leave it like this

go/exorelay/main.go
charlierudolph: no. I'll convert it to an int.

charlierudolph:
merged in 25dbacf

PR #121, opened 2017-05-23
go exoservice: reply to messages

charlierudolph:


go/exorelay/main.go
```diff
@@ -20,30 +20,33 @@ type Config struct {
...
-	return &ExoRelay{
-		config:         config,
-		messageChannel: make(chan structs.Message),
-	}
-}
-
 // Connect brings an ExoRelay instance online
 func (e *ExoRelay) Connect() error {
-	socket, err := websocket.Dial("ws://"+e.config.Host+":"+strconv.Itoa(e.config.Port), "", "origin:")
+	socket, err := websocket.Dial("ws://"+e.Config.Host+":"+strconv.Itoa(e.Config.Port), "", "origin:")
```
alexdavid: might be easier to use a `sprintf` here rather than `strconv`.

However after talking with @hugobho today, I realized since we're running these services in docker locally for development, we can actually always run exocom on 80/443 which would mean not having to specify the port here. This seems more idiomatic since the port will be inferred from the protocol (`ws` or `wss`).

What are your thoughts @charlierudolph @kevgo?

go/exoservice/exoservice.go
```diff
@@ -1 +1,77 @@
...
+
+func (e *ExoService) buildSendHelper(responseTo string) func(exorelay.MessageOptions) {
+	return func(options exorelay.MessageOptions) {
+		_, err := e.exoRelay.Send(exorelay.MessageOptions{
+			Name:       options.Name,
+			Payload:    options.Payload,
+			ResponseTo: responseTo,
+		})
+		if err != nil {
+			fmt.Println("Error sending message:", err)
```
alexdavid: I'm not sure if quietly logging is what we want to do here. It's probably fine for now, but we should add a TODO or a github issue to update this.

I feel like either the send should be handled by the caller or the service should crash. What do you think?

go/exoservice/exoservice.go
```diff
@@ -1 +1,77 @@
...
+	}
+	e.listenForMessages()
+	return nil
+}
+
+// Helpers
+
+func (e *ExoService) listenForMessages() {
+	messageChannel := e.exoRelay.GetMessageChannel()
+	go func() {
```
alexdavid: I think this function should block. the `listenForMessages` function in exorelay blocks below and so does `http.ListenAndServe`. I think it's reasonable if you want the exposed `Connect` function to not block, but then we should break out the goroutine there.

```go
go e.listenForMessages()
```

charlierudolph: sounds good

go/exoservice/exoservice.go
charlierudolph: I'm good with having this err.

go/exorelay/main.go
charlierudolph: The host currently doesn't contain ws / wss. That should be dealt with separately either way.

go/exorelay/exorelay_test.go
```diff
@@ -62,11 +62,13 @@ func FeatureContext(s *godog.Suite) {
 	})
 
 	s.Step(`^an ExoRelay with the role "([^"]*)"$`, func(role string) error {
-		exoInstance = exorelay.New(exorelay.Config{
-			Host: "localhost",
-			Port: port,
-			Role: role,
-		})
+		exoInstance = &exorelay.ExoRelay{
```
kevgo: What about creating a Go-like constructor?

```go
exoInstance = exorelay.New(exorelay.Config{...})

File: go/exoservice/test-fixtures/ping.go [hunk 5]
```diff
@@ -0,0 +1,29 @@
...
+type PingTextFixture struct {
+	ReceivedMessages []structs.Message
+}
+
+// Setup creates an ExoService instance for the given config
+func (r *PingTextFixture) Setup(config exorelay.Config) *exoservice.ExoService {
+	return &exoservice.ExoService{
+		Config: config,
+		MessageHandlers: exoservice.MessageHandlerMapping{
+			"ping": func(_ structs.MessagePayload, helpers exoservice.SendHelpers) {
```
kevgo: To simplify the method arguments, we could just provide a single object here. This is the pattern that https://github.com/gin-gonic/gin and many other frameworks use:

```go
"ping": func(r exoservice.Request) {
  // request payload is r.Payload
  err := r.Reply(exorelay.MessageOptions{Name: "pong"})
  ...

PR #122, opened 2017-05-24
exorelay: fix test inconsistency

charlierudolph:
message-id is used in the gherkin because its what `exocom.send` wants. I like using just `id` in the tests as then the block looks like the json of the actual message. I'm good updating the interface of `exocom.send` if that would be preferable


charlierudolph:
@kevgo I'm confused which part of the code you are referring to.

The actual json for the message uses [id](https://github.com/Originate/exocom/blob/25dbacfb39dd5f045f7aa9d137c0725aecbe6144/javascript/exocom-mock/src/mock-exo-com.ls#L49). Its just the interface to the send method of exocom-mock that expects message-id instead of id.

I wanted this update in order to make the gherkin contain the json of the message sent.

alexdavid:
This change makes sense to me if `id` is really what is being transmitted over the wire. I think it would be better if we keep the wire protocol in line with the developer API.

Also `id` isn't a bad name even if later we introduce request id, etc... because it is on a message the fact it is an id of the message.

PR #123, opened 2017-05-24
go exoservice: send messages

charlierudolph:


go/exoservice/test-fixtures/send.go
```diff
@@ -0,0 +1,29 @@
...
+
+// Setup creates an ExoService instance for the given config
+func (r *SendTextFixture) Setup(config exorelay.Config) *exoservice.ExoService {
+	return &exoservice.ExoService{
+		Config: config,
+		MessageHandlers: exoservice.MessageHandlerMapping{
+			"ping": func(_ structs.MessagePayload, helpers exoservice.SendHelpers) {
+				err := helpers.Send(exorelay.MessageOptions{Name: "ping_received"})
+				if err != nil {
+					fmt.Println("Failed to send reply")
```
alexdavid: Can we make this fail the test if it gets called?

PR #124, opened 2017-05-24
WIP: Changing exocom default port to 80

ben-goldberg:
would resolve https://github.com/Originate/exosphere-sdk/issues/276

The issue asks to change the default port for exocom to 80, since it'll be running inside a docker container anyways and shouldn't need to worry about if that port is already taken or not. 

Problem is, when running the tests locally, using port 80 requires privileged access. That is, they fail normally, but pass with `sudo`. 

Thoughts? @hugobho @kevgo @alexdavid


hugobho:
Just to be clear, using port 80 works in the Docker VM, maybe because we're automatically the root user there. There's only an issue when running tests locally. 

We wanted to use port 80 because it was the default non-ssl port, but I guess we could easily use some other default port (like 3100 haha). This PR still wouldn't be irrelevant though, we'd still have to make the necessary changes to ExoRelay.

charlierudolph:
For the tests, could they be updated to never use the default port?

ben-goldberg:
I mean, yes I could remove the tests around using the default port. But then, we wouldn't be testing that it works with the default port?

alexdavid:
Wait I thought we were testing this in docker as well. If there are cases where exocom still runs locally then we shouldn't do this change.

However, I'm curious, why aren't we running the end-to-end tests in docker since it will basically simulate more closely to how it will be running in production

hugobho:
Yeah, I think that's a good idea to test Exocom running in a Docker container. I still like having ExoRelay use a default port for now, and then we can switch them to port 80 once we start running tests in Docker.

javascript/exorelay/src/exorelay.ls
```diff
@@ -11,8 +11,8 @@ class ExoRelay extends EventEmitter
 
   (@config) ->
     @config?.exocom-host or throw new Error 'ExoCom host not provided to Exorelay'
-    @config?.exocom-port or throw new Error 'ExoCom port not provided to Exorelay'
     @config?.role or throw new Error 'Role not provided to Exorelay'
+    @config?.exocom-port or 80
```
kevgo: Should this be changed to `or=`?

charlierudolph:
Closing as the issue is suppose to resolve was decided against

PR #131, opened 2017-06-06
go exocom-mock: wait for message with name

charlierudolph:


charlierudolph:
Todo dry up the duplcation some code into utils

wgottschalk:
Overall the change looks fine to me, but as someone who is new on the project, I'm wondering what motivated the change.

charlierudolph:
In attempting to write a service using the go driver, I ran into a need for a this. Also just felt like a better way as this is what the `waitForReceivedMessageCount` was being used to accomplish in most cases.

PR #132, opened 2017-06-07
go: document setup

charlierudolph:


charlierudolph:
@ben-goldberg the services shouldn't actually import each (adding a copy in the vendor folder). I believe glide will sometimes do this (when updating). I added in some ignore statements to prevent this in the future.

go/README.md
```diff
@@ -0,0 +1,11 @@
+# Development
+
+The repo needs to live in a particular place in order for go to find everything properly
+
+* set the environment variable `$GOPATH` to your Go workspace (typically `~/go`)
+* create the directory `$GOPATH/src/github.com/Originate`
+* in that directory, clone this repository
```
charlierudolph: commands added

PR #133, opened 2017-06-07
Add glide.yaml  and vendor folder to top level

wgottschalk:
resolves https://github.com/Originate/exosphere-sdk/issues/285
WIP 
the plan is to aggregate all of the `glide.yaml` files to a top level file. Then delete the glide files for each subdirectory.
Not sure what to do with the package statements though. Does it change to just 
`github.com/Originate/exocom/go`?


go/glide.yaml
```diff
@@ -0,0 +1,14 @@
+package: github.com/Originate/exocom/go
+import:
+- package: github.com/fatih/color
+  version: v1.1.0
+- package: golang.org/x/net
+  subpackages:
+  - websocket
+- package: gopkg.in/tylerb/graceful.v1
+  version: v1.2.13
```
charlierudolph: `github.com/fatih/color` and `gopkg.in/tylerb/graceful.v1` are unused and should be removed.

charlierudolph:
I think the `bin/setup` command in each folder should be combined into a `bin/setup` in the go directory. The top level `bin/setup_ci` will need to be updated to just call that instead of calling instead of `../morula all bin/setup`

wgottschalk:
@charlierudolph @kevgo ready for final review. Lots of files changed due to moving vendor folders but the important ones should be at the top

go/bin/lint
```diff
@@ -3,5 +3,5 @@ set -e
 
 # required by gotype to work properly
 # see https://github.com/golang/go/issues/12703#issuecomment-144429128
-go install
-gometalinter.v1 --deadline 60s ./ ./test-fixtures
+go install $(go list ./... | grep -v /vendor/)
+gometalinter.v1 --deadline 60s ./
```
charlierudolph: gometalinter is not recursive so you need to specify the directories you want to lint

go/bin/spec
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -e
+
+bin/lint
+
+# runs go test on all subdirectories exluding the vendor folder
+# might cause issues with windows users though
```
charlierudolph: Why might this cause issues with windows users?

go/exorelay/exorelay_test.go
```diff
@@ -73,6 +74,7 @@ func FeatureContext(s *godog.Suite) {
 	})
 
 	s.Step(`^ExoRelay connects to Exocom$`, func() error {
+		time.Sleep(time.Millisecond * 1)
```
charlierudolph: Thoughts on making connect retry on failure? (with a timeout of 1 second)

go/bin/spec
wgottschalk: can windows users run bash commands?

go/exorelay/exorelay_test.go
wgottschalk: I think it's not in the scope of this PR but it's something that we should be doing

charlierudolph: K. Created an issue for it https://github.com/Originate/exosphere-sdk/issues/302

go/bin/spec
charlierudolph: http://gnuwin32.sourceforge.net/packages/grep.htm
How about updating the comment to: `(windows users need to 'grep' installed)`

charlierudolph: Actually thoughts on moving all our code into `go/src/` so that we don't need to filter out the vendor folder?

wgottschalk: I was actually looking at morula and it looks like there is a never command to exclude certain directories. So I could just say `cd go && ../morula all --never=vendor bin/<command>` in the top bin folder.

charlierudolph: That sounds good. Will also need to exclude the `bin` folder if we are keeping the shared setup

PR #134, opened 2017-06-08
WIP WebsocketBridge

ben-goldberg:
resolves https://github.com/Originate/exosphere-sdk/issues/291

This is still definitely a WIP, as I think I've broken tests in other places, and this is still very messy. But the PR is growing large so I wanted to get some feedback.

This creates a Websocket Bridge that acts as the point of communication between a browser and Exocom. To keep track of users, it creates a SessionID for each user. Then, it essentially just takes messages from a client (a browser) and forwards them to Exocom, and it takes messages from Exocom and forwards them to the right client.

Let me know what you think so far.

@charlierudolph @hugobho @kevgo


go/websocketbridge/features/clientToServer.feature
```diff
@@ -0,0 +1,30 @@
+Feature: 
+
+  As an Exosphere developer
+  I want to be able to send messages from a web client to my Exosphere application
+  So that I can create a useful application
+
+  Rules
+  - If a message does not have a SessionId, a new one is generated for that user
```
charlierudolph:
```
- A new session id is created for each websocket connection
```
Would be good to add a test that the if a second client connects, it has a different session id then the first

go/websocketbridge/main.go
```diff
@@ -0,0 +1,133 @@
...
+			if err != nil {
+				break
+			}
+			err = websocket.Message.Send(socket, serializedBytes)
+			if err != nil {
+				break
+			}
+		}
+	}()
+	return nil
```
charlierudolph: Would be nice to break this method up. The second part could become a private method, `listenForMessagesFromExorelay` and this methods calls it with
```golang
go w.listenForMessagesFromExorelay()
```

go/websocketbridge/main.go
```diff
@@ -0,0 +1,133 @@
...
+		return err
+	}
+	messageChannel := w.exoRelay.GetMessageChannel()
+	go func() {
+		for {
+			message, ok := <-messageChannel
+			if !ok {
+				break // channel closed
+			}
+			socket := w.SocketMap[message.SessionID]
```
charlierudolph: We should probably check if the socket exists here. This could happen if a client connects, sends a message and then disconnects before a reply arrives.

go/websocketbridge/main.go
```diff
@@ -0,0 +1,133 @@
...
+			message.SessionID = ""
+			serializedBytes, err := json.Marshal(message)
+			if err != nil {
+				break
```
charlierudolph: I don't think we should break here. Just because unmarshalling a message goes wrong, we shouldn't drop connection. I'd suggest just outputting the error with something like:
```
fmt.Printf("Error serializing message: %s", err.Error())
```
Same for the failing to send a message to a socket. That error message could include the sessionId and the error encountered

go/websocketbridge/main.go
```diff
@@ -0,0 +1,133 @@
...
+				return err
+			}
+		}
+	}
+	w.SocketMap = nil
+	return nil
+}
+
+func (w *WebsocketBridge) websocketHandler(socket *websocket.Conn) {
+	var sessionID string
```
charlierudolph: I'd suggest creating the sessionId and updating the socketMap here. I don't think we need to wait for the the client to send a message to do that.

go/websocketbridge/features/serverToClient.feature
```diff
@@ -0,0 +1,37 @@
+Feature: 
+
+  As an Exosphere developer
+  I want to be able to send messages from my Exosphere application to a client
+  So that I can create a useful application
+
+  Rules
+  -
```
charlierudolph:
```
- messages received from exorelay are forwarded to the client if and only if the session id matches
```
Couple tests to add here would be: 
* a message is received with no session id, ensure it isn't sent to any clients
* a message is received with a session id, check it isn't sent to other clients

go/websocketbridge/main.go
```diff
@@ -0,0 +1,133 @@
...
+
+	"golang.org/x/net/websocket"
+)
+
+// Config contains the configuration values for ExoRelay
+type Config struct {
+	Host string
+	Port int
+	Role string
+}
```
charlierudolph: We can just use the exorelay.Config instead of creating a config with the same structure

go/websocketbridge/main.go
```diff
@@ -0,0 +1,133 @@
...
+
+// WebsocketBridge handles communication between an Exosphere backend and
+// a browser front-end using websockets
+type WebsocketBridge struct {
+	ClientPort int
+	Config     Config
```
charlierudolph: I don't think `Config` or `ClientPort` needs to be saved on the Websocket bridge. `Config` was be passed into the `ConnectToExocom` method and `clientPort` can be passed into the `ListenForClients` methods.

It might also be a good idea to just create a single `Open` method which takes in both as I don't think we will want to one without the other

go/websocketbridge/main.go
```diff
@@ -0,0 +1,133 @@
...
+	w.server = http.Server{
+		Handler: websocket.Handler(w.websocketHandler),
+		Addr:    fmt.Sprintf(":%d", w.ClientPort),
+	}
+	go w.server.ListenAndServe()
+	return nil
+}
+
+// Reset closes all the open sockets and nils the SocketMap
+func (w *WebsocketBridge) Reset() error {
```
charlierudolph: I don't think we will ever actually Reset a websocket bridge in production. I'd suggest creating a close instead. For the tests we can close created websocket bridges and create a new one for each test.

go/websocketbridge/main.go
```diff
@@ -0,0 +1,133 @@
...
+	return nil
+}
+
+func (w *WebsocketBridge) websocketHandler(socket *websocket.Conn) {
+	var sessionID string
+	for {
+		message, err := w.readClientMessage(socket)
+		if err != nil {
+			fmt.Println("Error reading message from websocket:", err)
+			break
```
charlierudolph: An error here doesn't always need to be outputted. Please see exorelay as if the err is EOF, then the websocket has been closed.

After we break out the for loop, we should make sure we close the websocket and update the websocket map.

go/websocketbridge/test-fixtures/index.go
```diff
@@ -0,0 +1,26 @@
+package exorelayTestFixtures
```
charlierudolph: I believe this can be removed as the text fixtures aren't used

go/websocketbridge/websocket_bridge_test.go
```diff
@@ -0,0 +1,202 @@
...
+	})
+
+	s.Step(`websocket bridge makes the websocket request:`, func(payloadStr *gherkin.DocString) error {
+		t := template.New("request")
+		t, err := t.Parse(payloadStr.Content)
+		if err != nil {
+			return err
+		}
+		var expectedMessageBuffer bytes.Buffer
+		err = t.Execute(&expectedMessageBuffer, map[string]interface{}{"outgoingMessageId": outgoingMessageId})
```
charlierudolph: You don't appear to be using the template / outgoingMessageId logic, so it can be removed.

go/websocketbridge/websocket_bridge_test.go
```diff
@@ -0,0 +1,202 @@
...
+		return nil
+	})
+
+	s.Step(`receiving this message from the client:`, func(payloadStr *gherkin.DocString) error {
+		origin := "http://localhost/"
+		url := fmt.Sprintf("ws://localhost:%v/ws", clientPort)
+		var err error
+		ws, err = websocket.Dial(url, "", origin)
+		if err != nil {
+			log.Fatal(err)
```
charlierudolph: Please return the err instead of using `log.Fatal`. Step definitions can return an error which godog would report instead of this killing the suite

go/websocketbridge/websocket_bridge_test.go
```diff
@@ -0,0 +1,202 @@
...
+		err = json.Unmarshal(expectedMessageBuffer.Bytes(), &expectedMessage)
+		if err != nil {
+			return err
+		}
+		actualMessage, err := exocom.WaitForMessageWithName(expectedMessage.Name)
+		if err != nil {
+			return err
+		}
+		expectedMessage.ID = actualMessage.ID //ID gets generated on the fly, so we can't know it beforehand
+		if expectedMessage.SessionID == "" {
```
charlierudolph: I don't think this if guard is needed. We can just set the expectedMessage session id to that of actualMessage no matter what

go/websocketbridge/websocket_bridge_test.go
```diff
@@ -0,0 +1,202 @@
...
+	}()
+	return exocom
+}
+
+// Cucumber step definitions
+// nolint gocyclo
+func FeatureContext(s *godog.Suite) {
+	var exocom *exocomMock.ExoComMock
+	var websocketBridgeInstance *websocketbridge.WebsocketBridge
+	var ws *websocket.Conn
```
charlierudolph: can we please name this `clientWebsocket` instead of `ws` so its purpose is more clear?

go/websocketbridge/main.go
ben-goldberg: That's what I'm actually doing, I'll rename this to `Close` since the `reset` name gives the idea the same object will be used again.

ben-goldberg:
@charlierudolph comments addressed!

go/websocketbridge/bin/spec_ci
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+set -e
+
+bin/spec
```
ben-goldberg: I only see 4 comments and they all say "same as above"?

go/websocketbridge/LICENSE
```diff
@@ -0,0 +1,20 @@
+Copyright (c) 2016 Alec Snyder
```
charlierudolph: You aren't @allonsy. 😄

go/websocketbridge/websocket_bridge_test.go
```diff
@@ -0,0 +1,233 @@
...
+	})
+
+	s.Step(`a websocket bridge connected to Exocom`, func() error {
+		websocketBridgeInstance = &websocketbridge.WebsocketBridge{}
+		config := exorelay.Config{Host: "localhost", Port: exoPort, Role: "websocket-test"}
+		err := websocketBridgeInstance.Open(config, clientPort)
+		if err != nil {
+			return err
+		}
+		return nil
```
charlierudolph: these last five lines can be simplified to 
```go
return websocketBridgeInstance.Open(config, clientPort)
```

go/websocketbridge/websocket_bridge_test.go
```diff
@@ -0,0 +1,233 @@
...
+			marshaledMessage, err := json.Marshal(message)
+			if err != nil {
+				return err
+			}
+			_, err = clientWebsocket.Write(marshaledMessage)
+			if err != nil {
+				return err
+			}
+		}
+		return err
```
charlierudolph: You can return nil here as we are exiting directly if we ever hit an err.

go/websocketbridge/websocket_bridge_test.go
```diff
@@ -0,0 +1,233 @@
...
+			return fmt.Errorf("Expected request to equal %s but got %s", expectedMessage, actualMessage)
+		}
+		return nil
+	})
+
+	s.Step(`^the websocket bridge does not send a message to the client$`, func() error {
+		var bytes []byte
+		clientWebsocket.SetReadDeadline(time.Now().Add(time.Second * time.Duration(1)))
+		err := websocket.Message.Receive(clientWebsocket, &bytes)
+		if strings.Contains(err.Error(), "i/o timeout") {
```
charlierudolph: We should check if err is nil before calling `err.Error()`

go/websocketbridge/websocket_bridge_test.go
```diff
@@ -0,0 +1,233 @@
...
+				var actualMessage structs.Message
+				err = json.Unmarshal(bytes, &actualMessage)
+				if err != nil {
+					return err
+				}
+				return fmt.Errorf("Expected no message, but received %v", actualMessage)
+			}
+			return nil
+		} else {
+			fmt.Printf(err.Error())
```
charlierudolph: I don't think we need to print the error here since we are returning it, which would cause godog to print it.

go/websocketbridge/main.go
```diff
@@ -0,0 +1,150 @@
...
+		if err == io.EOF {
+			break
+		} else if err != nil {
+			fmt.Println("Error reading message from websocket:", err)
+			break
+		} else {
+			_, err := w.exoRelay.Send(exorelay.MessageOptions{Name: message.Name,
+				Payload:    message.Payload,
+				ResponseTo: message.ResponseTo,
+				SessionID:  sessionID})
```
charlierudolph: Can we format this in the following way:
```go
_, err := w.exoRelay.Send(exorelay.MessageOptions{
  Name: message.Name,
  Payload:    message.Payload,
  ResponseTo: message.ResponseTo,
  SessionID:  sessionID
})
```

go/websocketbridge/main.go
```diff
@@ -0,0 +1,150 @@
...
+
+// ConnectToExocom sets up an ExoRelay to communicate with exocom
+func (w *WebsocketBridge) connectToExocom(config exorelay.Config) error {
+	w.exoRelay = exorelay.ExoRelay{
+		Config: exorelay.Config{
+			Host: config.Host,
+			Port: config.Port,
+			Role: config.Role,
+		},
+	}
```
charlierudolph: I believe this can be simplifed to
```go
w.exoRelay = exorelay.ExoRelay{Config: config}
```

go/websocketbridge/websocket_bridge_test.go
```diff
@@ -0,0 +1,233 @@
...
+		}
+		err = exocom.WaitForConnection()
+		if err != nil {
+			return err
+		}
+		if message.SessionID != "" {
+			for key := range websocketBridgeInstance.SocketMap {
+				message.SessionID = key
+			}
+		}
```
charlierudolph: I'm not a fan of what this is doing. Can we set the session id based on a message sent to exocom? I'd prefer to keep the socket map private

go/websocketbridge/websocket_bridge_test.go
```diff
@@ -0,0 +1,233 @@
...
+		return exocom.Send(message)
+	})
+
+	s.Step(`^those clients have different sessionIds$`, func() error {
+		if len(websocketBridgeInstance.SocketMap) != clientCount {
+			return fmt.Errorf("Expected request to have %d clients but got %d", clientCount, len(websocketBridgeInstance.SocketMap))
+		}
+		return nil
```
charlierudolph: Can we test this step by having each client send a message to exocom and checking that the messages have different session ids? I'd prefer to keep the socket map private

go/websocketbridge/websocket_bridge_test.go
ben-goldberg: oops! that was a debugging print statement I forgot to remove

ben-goldberg:

go/websocketbridge/bin/spec_ci
wgottschalk: yeah nvm with these. You still keep these files with with my changes. The only change is that `vendor` folder and `glide.yaml` are now in the top level so you won't need a local vendor folder anymore if you already have one.

charlierudolph:
blocked on #145

PR #137, opened 2017-06-09
Moved individual vendors folds to a shared vendor folder

wgottschalk:
resolves Originate/exosphere-sdk#285

closing https://github.com/Originate/exocom/pull/133 in favor of this PR.
I didn't want to have to go through and re-add missing bash scripts so here's a fresh PR.

It's a lot simpler now. I moved vendor to /go and told morula to exclude running commands in the `vendor subdirectory`
@charlierudolph


charlierudolph:
glide install should be removed from each of bin/setup since thats not used in those directories anymore

PR #138, opened 2017-06-10
Adding retry to exorelay go

wgottschalk:
resolves https://github.com/Originate/exosphere-sdk/issues/302
Just posting this to have eyes on it as I work on the PR

@charlierudolph
@hugobho 
@ben-goldberg


wgottschalk:
@charlierudolph @kevgo let me know what you think of the PR. 
Right now, the Connect method on exorelay blocks until it connects to exocom which makes sense to me for the time being.

go/exorelay/features/retry.feature
```diff
@@ -0,0 +1,20 @@
...
+  Background:
+    Given an ExoRelay with the role "foo"
+
+  Scenario: ExoRelay attempts to establish a connection to ExoCom
+    Given Exocom is not online yet
+    When ExoRelay connects to Exocom while Exocom is booting up
+    Then it registers by sending the message "exocom.register-service" with payload:
+      """
+      { "clientName": "foo"} 
+      """
```
kevgo: Please also add a test where the connection is established, then disconnected, and Exorelay re-connects.

wgottschalk:
So after a while of trying to figure this out, it looks like the go standard websocket lib doesn't provide a very nice way of handling closed connections. The gorilla websocket library does:
https://github.com/gorilla/websocket/blob/master/conn.go#L259

The issue is that unlike JS, Go's websockets are not event emitters so we can't just listen for changes and react. I cannot figure out a way to get the client side (exorelay) notified when the server (exocom) disconnects. 

My plan was to check for an error on the exorelay side after sending a message and try to reconnect from there but that doesn't seem to be working for some reason.

kevgo:
Using Gorilla-Websocket is definitely an option here.
If we can't figure this out at all, the service should at least crash so that our runtime can restart it and it connects to the new ExoCom instance.

charlierudolph:
closing in favor of #145

PR #140, opened 2017-06-13
go: enforce formatting

charlierudolph:
We used this on git-town as it enforces that `go fmt` is run on the code. Editors should be configured to fmt the code on save


wgottschalk:
I've been using goimports instead of gofmt for formatting. Perhaps it's worth making the switch to enforce clean imports as well?

kevgo:
+1 if goimports has a similar command-line switch

charlierudolph:
Good idea. done

PR #141, opened 2017-06-19
Check subdirs are same version in publish script

hugobho:
Resolves https://github.com/Originate/exosphere-sdk/issues/253


bin/publish
```diff
@@ -51,6 +54,19 @@ function main
   push-git-tags!
 
 
+function verify-subproject-versions
+  console.log green "Verifying subproject version numbers...\n"
+  subproject-versions = {}
+  subproject-versions[exec("npm view exocom-dev version")] = null
+  subproject-versions[exec("npm view exocom-mock version")] = null
+  subproject-versions[exec("npm view exorelay version")] = null
+  subproject-versions[exec("npm view exoservice version")] = null
```
charlierudolph: do we need to run `npm view` here? Could we just require the `package.json` and look at the version?

hugobho:

PR #143, opened 2017-06-20
forward session ID

cheahuychou:
resolves: https://github.com/Originate/exosphere-sdk/issues/292

Attach session ID to each message


exocom-server/features/support/world.ls
```diff
@@ -116,7 +116,19 @@ World = !->
...
+
+  @verify-session-id = ({client-name, message-name, id = '123', response-to}, done) ~>
+    service = @service-mocks[client-name]
+    wait-until (~> service.received-messages[0]?.name && service.received-messages.length == 2), 1, ~>
+      service.received-messages.for-each (message) -> expect(message.session-id).to.match /^\w|-*$/
+      expect(service.received-messages[0].session-id).to.eql(service.received-messages[1].session-id)
+      if @session-id
+        expect(@session-id).to.not.eql(service.received-messages[0].session-id)
+      @session-id = service.received-messages[0].session-id
+      done!
```
hugobho: Let's extract this logic into two separate functions: one that tests that the array has equal session ids and another that checks a new session id has been generated.

cheahuychou:
@hugobho all comments addressed.

exocom-server/features/support/steps-then.ls
```diff
@@ -14,6 +14,15 @@ module.exports = ->
     @verify-sent-calls {client-name, message-name, response-to: '111'}, done
 
 
+  @Then /^ExoCom broadcasts replies to the "([^"]+)" service with the same session ID$/, (client-name, done) ->
+    @verify-equal-session-id {client-name, "users.created", response-to: '111'}, done
+
+
+  @Then /^ExoCom broadcasts replies to the "([^"]+)" service with a new session ID$/, (client-name, done) ->
+    @verify-equal-session-id {client-name, "users.created", response-to: '111'}, ~>
+      @verify-new-session-id {client-name, "users.created", response-to: '111'}, done
```
hugobho: Does it make a little more sense if if we `@verify-new-session-id` first and then `@verify-equal-session-id`?

exocom-server/features/support/steps-when.ls
```diff
@@ -1,4 +1,5 @@
 require! {
+  'async'
```
hugobho: Is this being used?

exocom-server/features/support/world.ls
```diff
@@ -116,7 +116,26 @@ World = !->
...
+      expect(received-message).to.include expected
+      expect(received-message.session-id).to.match /^\w|-*$/
+      done!
+
+
+  @verify-equal-session-id = ({client-name, message-name, id = '123', response-to}, done) ~>
+    service = @service-mocks[client-name]
+    wait-until (~> service.received-messages[0]?.name && service.received-messages.length == 2), ~>
+      service.received-messages.for-each (message) -> expect(message.session-id).to.match /^\w|-*$/
+      expect(service.received-messages[0].session-id).to.eql(service.received-messages[1].session-id)
```
hugobho: So that this method is more generic, let's check that all elements in the array have the same value. We could check that `every` element has the same value as the first one: https://lodash.com/docs/4.17.4#every

exocom-server/features/support/world.ls
```diff
@@ -116,7 +116,26 @@ World = !->
...
       expected.response-to = that if response-to
-      jsdiff-console received-message, expected, done
+      expect(received-message).to.include expected
+      expect(received-message.session-id).to.match /^\w|-*$/
+      done!
+
+
+  @verify-equal-session-id = ({client-name, message-name, id = '123', response-to}, done) ~>
+    service = @service-mocks[client-name]
+    wait-until (~> service.received-messages[0]?.name && service.received-messages.length == 2), ~>
```
hugobho: This (`service.received-messages.length == 2`) is tied a little too closely to some of the other step definitions. Let's do `service.received-messages.length > 1`. Also, why are we checking for `.name` here?

exocom-server/features/support/world.ls
hugobho: We should be setting `@session-id` here as well right? If that variable's going to continuously keep track of the latest session.

exocom-server/features/support/world.ls
```diff
@@ -116,7 +116,26 @@ World = !->
...
+      service.received-messages.for-each (message) -> expect(message.session-id).to.match /^\w|-*$/
+      expect(service.received-messages[0].session-id).to.eql(service.received-messages[1].session-id)
+      done!
+
+
+  @verify-new-session-id = ({client-name, message-name, id = '123', response-to}, done) ~>
+    service = @service-mocks[client-name]
+    wait-until (~> service.received-messages[0]?.name && service.received-messages.length == 2), ~>
```
hugobho: Why `service.received-messages[0]?.name` here as well? Let's also make the conditional `service.received-messages.length > 1`

exocom-server/features/support/world.ls
```diff
@@ -116,7 +116,26 @@ World = !->
...
+      if @session-id
+        expect(@session-id).to.not.eql(service.received-messages[0].session-id)
```
hugobho: Let's use `lodash/every` here to check against the whole array as well.

exocom-server/features/support/world.ls
cheahuychou: I used `for-each` just to be consistent with the previous line. I also find it easier to read.

exocom-server/features/support/world.ls
cheahuychou: I meant to check that it is equal to `message-name` but accidentally got rid of that while debugging. I've put it back.

exocom-server/src/websocket-subsystem/websocket-subsystem.ls
```diff
@@ -65,6 +68,7 @@ class WebSocketSubsystem extends EventEmitter
   # Registers the given websocket as a connection
   # to an instance of the service with the given name
   register-client: ({client-name, websocket}) ->
+    @session-ids[client-name] = uuid.v1!
```
charlierudolph: Session ids will be passed in the message and are created only on the websocket bridge. Exocom should not be creating them. I had envisioned this PR only touching the exorelay implementations.

charlierudolph:
Closing since this was the wrong implementation. I'll add some comments to the original issue to hopefully make it more clear when we tackle this in the future

PR #146, opened 2017-06-30
Exocom: do not exit when sending unregistered message

hugobho:
Exocom should handle unregistered messages gracefully instead of throwing and error and exiting.


charlierudolph:
Can we add a test for this? It appears we already have tests checking the error message we would just need to add a line checking that this does not kill the process.

charlierudolph:
@hugobho any updates here? I can help out with this PR as well if you'd like

PR #145, opened 2017-06-29
exorelay reconnect to exocom

charlierudolph:
resolves https://github.com/Originate/exosphere-sdk/issues/302 

rewrite of #138 on top of #144

PR #158, opened 2017-07-06
javascript exorelay: reconnect if exocom goes down

charlierudolph:
replacement for #145 (I updated cucumber-js separately while trying to figure out what was going on)


javascript/exorelay/src/websocket-connector/websocket-connector.ls
```diff
@@ -56,20 +62,27 @@ class WebSocketConnector extends EventEmitter
...
+      @connect!
 
 
   _on-socket-open: ~>
     @emit 'online'
 
 
   _on-socket-error: (error) ~>
-    | error.errno is 'EADDRINUSE'   =>  @emit 'error', "port #{@exocom-port} is already in use"
-    | error.errno is 'ECONNREFUSED' =>  wait 1_000, @connect
```
charlierudolph: This line was unnecessary for the current tests to pass. It appears the websocket library we are using keeps retrying on the initial connections even though it emits an error.

javascript/exorelay/src/websocket-connector/websocket-connector.ls
```diff
@@ -24,11 +24,17 @@ class WebSocketConnector extends EventEmitter
...
   # Closes the port that ExoRelay is communicating on
-  close: ->
-    return unless @socket
+  close: (done) ->
+    return done! unless @socket
     debug "no longer connected at 'ws://#{@exocom-host}/#{@exocom-port}'"
-    @socket?.close!
-    @emit 'offline'
+    @should-reconnect-on-socket-closed = no
+    switch @socket.readyState
```
hugobho: `@socket.ready-state`

javascript/exorelay/src/websocket-connector/websocket-connector.ls
```diff
@@ -24,11 +24,17 @@ class WebSocketConnector extends EventEmitter
...
+      | WebSocket.CONNECTING => done!
```
hugobho: We don't need to call anything to terminate the socket here?

charlierudolph: Updated. Can't call close as that would error but using `terminate` works.

javascript/exorelay/src/websocket-connector/websocket-connector.ls
charlierudolph: fixed

kevgo:
I think `push` just checks the actual commit, and `pr` checks a different commit that is the current commit merged with `master`.

PR #159, opened 2017-07-07
go exorelay: stabilize tests

charlierudolph:
Getting some random failures as we have a test that was just waiting for a connection and killing the connection before the register service message could be sent was causing some null pointer errors.


go/exocom-mock/exocom-mock.go
```diff
@@ -61,31 +64,6 @@ func (e *ExoComMock) HasConnection() bool {
...
-	if e.HasConnection() {
-		err := e.socket.Close()
-		e.socket = nil
-		if err != nil {
-			return err
-		}
-	}
-	e.ReceivedMessages = []structs.Message{}
-	return nil
-}
```
charlierudolph: These methods were no longer needed as we are no longer resetting the mock anywhere. We are using a new mock in every test and in the retry test we are now closing the mock and opening another one

PR #161, opened 2017-07-12
exocom-server: remove responseTo in favor of activityId

charlierudolph:
Assumptions:
* Every message has an activityId. When sending a message it will be provide an activityId (when replying or adding an insight) or have on auto-generated similar to id (when starting a new discussion). 

Comments:
* The responseTime code had to be removed since it depended on responseTo and you cannot calculate it with activityId alone as you don't know which message in the activity it is responding to.
* message id appears to be unused now aside from being a unique identifier for each message


kevgo:
+1 on removing message id

I think we can still calculate responseTime. Take the latest message that was sent to the service instance with the same activityId as the baseline.

exocom-server/src/exocom.ls
```diff
@@ -91,13 +89,6 @@ class ExoCom extends EventEmitter
     return @logger.warning "No receivers for message '#{message-data.name}' registered" unless subscribers
     subscriber-names = [subscriber.client-name for subscriber in subscribers]
 
-    # calculate a message's response time if it is a reply
```
kevgo: Let's keep calculation of the reply time

charlierudolph:
Closing due to the rewrite in go. Will link to this PR in the issue to use as a template

PR #165, opened 2017-07-13
{DO NOT MERGE] exocom-server: rewrite in go

charlierudolph:
Just setting up the command and have the feature `configuring-the-ports.feature` passing. Had to delete one feature as it was erroring that the address was already in use (when trying to startup the mock service on the port). I'm planning on using a new port for each of the other tests.

I'll do each subsequent feature in another PR in hope of keeping this easy to review


go/exocom/bin/lint
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+set -e
+
+if [ -n "$(goimports -d . 2>&1)" ]; then
+  echo "Code is not formatted. Run 'goimports -w .'"
+  exit 1
+fi
+
+# required by gotype to work properly
```
kevgo: Since we disable `gotype` below, is this comment/line still necessary?

go/exocom/features/configuring-the-ports.feature
```diff
@@ -0,0 +1,24 @@
+Feature: configuring the port
+
+  As an Exosphere developer testing my application
+  I want to be able to boot ExoCom at various ports
```
kevgo: I don't think we need this feature anymore now that Exocom runs inside Docker. We can always run it at a certain port like 3100.

go/exocom/features/configuring-the-ports.feature
```diff
@@ -0,0 +1,24 @@
...
+  So that I have flexibility in testing my system.
+
+  - the default ExoCom port is 3100
+  - provide the "EXOCOM_PORT" environment variable to listen for HTTP requests on a custom port
+
+
+  Scenario: running at the default port
+    When I run ExoCom
```
kevgo: The terminology I found useful is:
- `When starting XXX` means the process is started, the step doesn't wait for anything, and the process is still running when the step finishes
- `When running XXX` means the process is started and the step waits until the process finished without an error code

There is also `When trying to run XXX`, which starts the process, waits until it finishes, and verifies that it exits with an error code.

go/exocom/features/configuring-the-ports.feature
```diff
@@ -0,0 +1,24 @@
...
+    Then it opens a port at 3100
```
kevgo: I would add a step defining the message Exocom prints to indicate that it is properly started up.

```
Then I see "Exocom is online at port 3100"

File: go/exocom/features/configuring-the-ports.feature [hunk 5]
```diff
@@ -0,0 +1,24 @@
...
+
+
+  Scenario: using a custom port
```
kevgo: This scenario can go away

go/exocom/features/configuring-the-ports.feature
```diff
@@ -0,0 +1,24 @@
...
+    When starting ExoCom at port 3200
+    Then it opens a port at 3200
+
+
+  Scenario: custom port that is already taken
```
kevgo: Scenario: port 3100 is already taken

go/exocom/src/cmd/index.go
```diff
@@ -0,0 +1,34 @@
...
+	Short: "",
+	Run: func(cmd *cobra.Command, args []string) {
+		port := os.Getenv("PORT")
+		if port == "" {
+			port = "3100"
+		}
+		fmt.Printf("ExoCom online at port %s\n", port)
+		err := http.ListenAndServe(fmt.Sprintf(":%s", port), nil)
+		if err != nil {
+			panic(err)
```
kevgo: Not sure if we want to panic here. Panics are pretty rare in Go. If there is an error here, let's exit cleanly using fmt.Fatal.

go/exocom/test_helpers/index.go
```diff
@@ -0,0 +1,48 @@
...
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/Originate/exocom/go/utils"
+)
+
+// WaitForText reads from the stream waiting for the given text
+func WaitForText(stream io.ReadCloser, text string) error {
```
kevgo: What do you think about extracting this into a dedicated Go library? We will need this on a ton of projects including Exosphere, Morula, etc.

kevgo: Does this have to be an `io.ReadCloser`? You seem to only need an `io.Reader` here.

go/utils/wait_for.go
```diff
@@ -11,6 +11,13 @@ import (
 // WaitFor waits a maximum of 10 seconds the given condition to become true
 // returning an error with the given message if and only if it does not
 func WaitFor(condition func() bool, errorMessage string) error {
+	return WaitForf(condition, func() error {
+		return errors.New(errorMessage)
+	})
+}
+
+// WaitForf is similar to WaitFor but accepts a func to generate the error message
+func WaitForf(condition func() bool, errorFn func() error) error {
```
kevgo: This is a very cool idea, feels pretty idiomatic!

go/utils/wait_for.go
```diff
@@ -27,7 +34,7 @@ func WaitFor(condition func() bool, errorMessage string) error {
 	case <-success:
 		return nil
 	case <-time.After(time.Second * 10):
-		return errors.New(errorMessage)
+		return errorFn()
```
kevgo: What about providing the error to the error generator function?

```go
return errorFn("Timeout")

charlierudolph:
Closing. Going to try and break this down into some passing PRs.

go/utils/wait_for.go
charlierudolph: error fn is only ever called during a timeout

go/exocom/test_helpers/index.go
charlierudolph: I think we have a better one over on exosphere now with our process_helpers

go/exocom/src/cmd/index.go
charlierudolph: panic includes the stack trace though. This is an unexpected error so I'd prefer to have the stack trace. Anytime where its an error we are planning for or know might fail, I think log.Fatal makes sense as we don't want to give the user the stack trace

go/exocom/features/configuring-the-ports.feature
charlierudolph: I found its better to use fresh ports in golang for tests as otherwise we get some errors that the port is still taken.

go/exocom/features/configuring-the-ports.feature
charlierudolph: I think its useful to have for tests

go/exocom/features/configuring-the-ports.feature
charlierudolph: Done

go/exocom/features/configuring-the-ports.feature
charlierudolph: Updated

go/exocom/features/configuring-the-ports.feature
charlierudolph: I think its useful to have for testing

go/exocom/bin/lint
charlierudolph: Enabled gotype and it found the ReadCloser vs Reader that you mentioned

charlierudolph:
@kevgo commented back as will address your comments before opening the new PR

PR #167, opened 2017-07-25
exocom - rewrite in go: configure the ports, subsystems, register-service

charlierudolph:
@kevgo responded to your comments from #165


go/exocom/main_test.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/DATA-DOG/godog"
+	"github.com/Originate/exocom/go/exocom/test_helpers"
+)
+
+// nolint gocyclo
+func FeatureContext(s *godog.Suite) {
+	var cmd *exec.Cmd
+	var cmdOutput string
+	var cmdStdout io.ReadCloser
+	var cmdStderr io.ReadCloser
```
cheahuychou: Thoughts on using a global `process` like [here](https://github.com/Originate/exosphere/blob/master/exo-go/test_helpers/shared_feature_context.go#L20) instead of using 4 global variables?

charlierudolph: Sounds good. I'll extract out process helpers from exosphere into a new repo and can use that in both places. I'll handle that in a separate PR though

PR #168, opened 2017-07-25
exocom - rewrite in go: subsystems

charlierudolph:
made for @mcharawi


go/exocom/src/client_registry/client_registry.go
```diff
@@ -0,0 +1,77 @@
+package clientRegistry
+
+import "encoding/json"
+
+// Client is the combination of a client name, service type and internal namespace
+type Client struct {
```
kevgo: Thoughts around extracting each struct into its own file with the struct name as the file name? That's how it is done in Ruby, and it's very helpful in finding implementations.

go/exocom/src/client_registry/client_registry.go
```diff
@@ -0,0 +1,77 @@
...
+	ClientName        string
```
kevgo: Didn't we name this `role`? The word "name" is too generic. There are three names here:
- service type: the "class" of the service
- instance name (coming soon): the name of the particular service instance ("user service instance #3")
- role: the role that the service instance plays inside the application (a MongoService instance could play the role UserService). This is only different when using message translation.

go/exocom/src/client_registry/client_registry.go
```diff
@@ -0,0 +1,77 @@
...
+	ServiceType       string
+	InternalNamespace string
+}
+
+// Route is an entry in the routes
+type Route struct {
```
kevgo: Shouldn't this be in some sort of subsystem, for example the broadcast subsystem?

go/exocom/src/client_registry/client_registry.go
```diff
@@ -0,0 +1,77 @@
...
+	Receives          []string
+	Sends             []string
+	InternalNamespace string
+}
+
+// ClientRegistry manages which clients are connected to exocom,
+// and what messages each client can send and receive
+type ClientRegistry struct {
+	Routing       map[string]Route
```
kevgo: this line should be in the broadcast subsystem

go/exocom/src/client_registry/client_registry.go
```diff
@@ -0,0 +1,77 @@
...
+	Clients       map[string]Client
```
kevgo: this map should be in the connection subsystem

go/exocom/src/client_registry/client_registry.go
```diff
@@ -0,0 +1,77 @@
...
+	subscriptions *SubscriptionManager
```
kevgo: This one probably also goes into the broadcast subsystem. Maybe add convenience methods to add and look up entries to it as well?

go/exocom/src/client_registry/client_registry_test.go
```diff
@@ -0,0 +1,106 @@
...
+		BeforeEach(func() {
+			serviceRoutes :=
+				`{
+					"role 1": {
+						"receives": ["message 3 name"],
+						"internalNamespace": "namespace"
+					}
+				}`
+			var err error
+			registry, err = clientRegistry.NewClientRegistry(serviceRoutes)
```
kevgo: Why are we providing the routes as a string here? That seems strange. I would expect we parse config files into Go structs.

go/exocom/src/client_registry/subscription_manager.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// SubscriptionManager manages what clients should be sent messages
+type SubscriptionManager struct {
+	MessageNameToSubscribers map[string][]Subscriber
+	Routing                  map[string]Route
+}
+
+// NewSubscriptionManager returns a SubscriptionManager for the given routes
+func NewSubscriptionManager(routes map[string]Route) *SubscriptionManager {
```
kevgo: Can we defined a `Routes` type that is `map[string]Route` to make the semantics of data more explicit and not juggle so many anonymous maps around?

go/exocom/src/message_cache/message_cache.go
```diff
@@ -0,0 +1,59 @@
...
+	return
+}
+
+// Get returns the timestamp for the given messageId returning an error if
+// no data is available
+func (c *MessageCache) Get(messageID string) (time.Time, error) {
+	c.mutex.Lock()
+	timestamp, ok := c.Cache[messageID]
+	if ok {
+		c.mutex.Unlock()
```
kevgo: I think you can move this line before the if clause. It is the direct next action in both casese.

go/exocom/src/message_cache/message_cache.go
```diff
@@ -0,0 +1,59 @@
...
+			delete(c.Cache, id)
+		}
+	}
+	return
+}
+
+// Get returns the timestamp for the given messageId returning an error if
+// no data is available
+func (c *MessageCache) Get(messageID string) (time.Time, error) {
+	c.mutex.Lock()
```
kevgo: Please add a comment why you are adding a lock here, it is not obvious from the code.

go/exocom/src/message_cache/message_cache.go
charlierudolph: This got refactored in a later PR where the branch was removed

go/exocom/src/message_cache/message_cache.go
charlierudolph: Sounds good. I'll do that in a separate PR

go/exocom/src/client_registry/subscription_manager.go
charlierudolph: Sounds good

go/exocom/src/client_registry/client_registry_test.go
charlierudolph: This is copying the node version of where the routes actually get parsed. We can move the parsing up the chain

go/exocom/src/client_registry/client_registry.go
charlierudolph: I'd prefer to not add those until we need them

go/exocom/src/client_registry/client_registry.go
charlierudolph: I believe its role in the service routes but was named clientName internally. Will update in a separate PR

go/exocom/src/client_registry/client_registry.go
charlierudolph: For the rewrite we just kept the same internals as in the js version. Let explore the subsystems when you get back as I'm unclear what exactly lies within each boundary

PR #172, opened 2017-07-27
JS - forward session ID

cheahuychou:
resolves: https://github.com/Originate/exosphere-sdk/issues/292


charlierudolph:
Please add new tests instead of updating existing ones.

PR #173, opened 2017-07-27
exocom - rewrite in go: broadcasting and translating messages

mcharawi:
@charlierudolph 
@cheahuychou 
@hugobho 
@kevgo 

Rewrite of exocom in Go.


go/structs/message.go
```diff
@@ -6,11 +6,13 @@ import "time"
 type Message struct {
 	ID           string         `json:"id"`
 	Name         string         `json:"name"`
+	OriginalName string         `json:"originalName"`
```
charlierudolph: I don't think we should have the `originalName` as part of the message struct. I'd like it to only define the fields that are sent over the websockets.

go/exocom/src/exocom/exocom.go
```diff
@@ -106,3 +125,71 @@ func parseRegisterMessagePayload(message structs.Message) (string, error) {
...
 	return "", fmt.Errorf("Invalid register message payload: %v", message.Payload)
 }
+
+func (e *ExoCom) send(message structs.Message) error {
+	publicMessageName := messageTranslator.GetPublicMessageName(&messageTranslator.GetPublicMessageNameOptions{
+		Namespace:           e.clientRegistry.Clients[message.Sender].InternalNamespace,
+		ClientName:          message.Sender,
+		InternalMessageName: message.Name,
+	})
+	message.OriginalName = message.Name
```
charlierudolph: I think we can get by without setting the OriginalName as a property on the message and just passing it a variable to `sendToServices`

go/exocom/main_test.go
```diff
@@ -153,10 +156,11 @@ func FeatureContext(s *godog.Suite) {
...
 		}
-		cmdStderr, err = cmd.StderrPipe()
-		if err != nil {
-			return err
-		}
+		cmd.Stderr = os.Stderr
+		//cmdStderr, err = cmd.StderrPipe()
+		//if err != nil {
+		//	return err
+		//}
```
charlierudolph: Please revert these changes

go/exocom/main_test.go
```diff
@@ -196,6 +215,40 @@ func FeatureContext(s *godog.Suite) {
 		}
 		return nil
 	})
+
+	s.Step(`^ExoCom broadcasts the message "([^"]*)" to the "([^"]*)" service$`, func(messageName, serviceName string) error {
+		//TODO: chekc that the received message matches the given message
```
charlierudolph: Is there anything todo here?

go/exocom/main_test.go
```diff
@@ -196,6 +215,40 @@ func FeatureContext(s *godog.Suite) {
...
+		_, err := services[serviceName].WaitForMessageWithName(messageName)
+		if err != nil {
+			return err
+		}
+		return nil
```
charlierudolph:
```go
if err != nil {
  return err
}
return nil

// can also be simplified to

return err

```

go/exocom/main_test.go
```diff
@@ -28,6 +29,8 @@ func FeatureContext(s *godog.Suite) {
 	var shouldCmdBeKilled bool
 	var server *http.Server
 	var exocomPort int
+	var lastSentMessage string
+	var lastSentMessageName string
```
charlierudolph: What are these variable used for? They appear to be assigned but are otherwise unused

go/exocom/main_test.go
```diff
@@ -196,6 +215,40 @@ func FeatureContext(s *godog.Suite) {
...
+	})
+
+	s.Step(`^ExoCom broadcasts the reply "([^"]*)" to the "([^"]*)" service$`, func(messageName, serviceName string) error {
+		//TODO: chekc that the received message matches the given message
+		_, err := services[serviceName].WaitForMessageWithName(messageName)
+		if err != nil {
+			return err
+		}
+		return nil
+	})
```
charlierudolph: This step definition is very similar to `ExoCom broadcasts the message ...`, since we aren't doing anything different I think we update the gherkin to just use one.

go/exocom/main_test.go
```diff
@@ -196,6 +215,40 @@ func FeatureContext(s *godog.Suite) {
...
+
+	s.Step(`^ExoCom outputs "([^"]*)"$`, func(errorText string) error {
+		return testHelpers.WaitForText(cmdStdout, errorText)
+	})
```
charlierudolph: Lets merge this and the `ExoCom signals ...` step definition by only using one in the gherkin

mcharawi:
comments addressed @charlierudolph

go/exocom/src/exocom/exocom.go
```diff
@@ -106,3 +125,71 @@ func parseRegisterMessagePayload(message structs.Message) (string, error) {
...
 }
+
+func (e *ExoCom) send(message structs.Message) error {
+	publicMessageName := messageTranslator.GetPublicMessageName(&messageTranslator.GetPublicMessageNameOptions{
+		Namespace:           e.clientRegistry.Clients[message.Sender].InternalNamespace,
+		ClientName:          message.Sender,
+		InternalMessageName: message.Name,
+	})
+	originalName := message.Name
+	message.Name = publicMessageName
```
charlierudolph: I think we can avoid overriding this as well. I'd like to try and not update message. I think we can get away with using a publicMessageName variable that gets passed into sendToServices

go/exocom/src/exocom/exocom.go
```diff
@@ -106,3 +125,71 @@ func parseRegisterMessagePayload(message structs.Message) (string, error) {
...
+	}
+	internalMessageNameMapping, err := e.sendToServices(message, subscribers)
+	if err != nil {
+		return err
+	}
+	printLogError(e.logger.Messages(message, originalName, internalMessageNameMapping))
+	return nil
+}
+
+func (e *ExoCom) sendToServices(message structs.Message, subscribers []clientRegistry.Subscriber) (map[string]string, error) {
```
kevgo: What sort of map does this method return? Ideally this is documented via a variable name for the result map.

go/exocom/test_helpers/mock_service.go
```diff
@@ -43,6 +47,15 @@ func (m *MockService) Connect() error {
 	return m.socket.WriteMessage(websocket.TextMessage, serializedBytes)
 }
 
+//Send llows the mock service to send a websocket message to exocom
```
kevgo: typo in `llows`

go/exocom/features/broadcasting-messages.feature
```diff
@@ -0,0 +1,66 @@
...
+  - ExoCom sends the message to all subscribed services
+
+
+  Background:
+    Given an ExoCom instance configured with the routes:
+    """
+    [
+      {
+        "role": "web",
+        "receives": ["users.created"],
```
mcharawi: we will address this when we change how message translation works.

PR #179, opened 2017-07-27
exocom: rename clientName to role

charlierudolph:
In response to https://github.com/Originate/exocom/pull/168#discussion_r129959325

Most changes are internal, except the output of requesting `config.json`

Related: the `exocom.registerService` message includes a payload `{clientName: "..."}`. Do you think that should be renamed as well? That should be updated in a separate PR as it needs to touch all the exorelay implementations


charlierudolph:
@kevgo any response to the questions in the PR description?

kevgo:
Yes, let's get all the vocabulary correct.

PR #185, opened 2017-08-01
update publish script

charlierudolph:
@kevgo @hugobho does docker need specific publishing rights?

I want to move publishing to docker / npm to travisci in the future so we don't need to hand out permissions.


go/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM golang:1.8
+
+WORKDIR /go/src/github.com/Originate/exocom/go/
+COPY . .
+
+WORKDIR exocom
+RUN go install
+RUN rm -rf /go/src/
+
+CMD /go/bin/exocom
```
charlierudolph: Copying the entire go folder to get the vendor and shared (util / struct) directories. Build the binary, then remove all source to make the docker image smaller. Finally give it the default command as bin/exocom and thus we don't need to supply a command on exosphere to run exocom

kevgo: What is this Dockerfile for? If for shipping Exocom, why can't we simply copy only the exocom binary into an empty Docker container? Or maybe one based on Busybox or Alpine Linux?

charlierudolph: That should work. I'm just not sure what version we need for the docker image and this way it build it on the machine it runs.

kevgo:
If we compile Exocom inside a Docker image, we should be able to simply
take the generated binary and put it into an empty Docker image.

charlierudolph:
This is now updated with the simpler publish process

PR #187, opened 2017-08-01
Mc response to activity

mcharawi:
remove responseTo in favor of activityId


go/exocom/src/exocom/exocom.go
```diff
@@ -140,14 +140,12 @@ func (e *ExoCom) send(message structs.Message) error {
...
 	}
-	if message.ResponseTo != "" {
-		originalTimestamp, ok := e.messageCache.Get(message.ID)
-		if ok {
-			message.ResponseTime = message.Timestamp.Sub(originalTimestamp)
-		}
-	} else {
-		e.messageCache.Set(message.ID, message.Timestamp)
+
+	originalTimestamp, ok := e.messageCache.Get(message.ID)
```
charlierudolph: We need to update the message cache to be a mapping from message.ActivityID to timestamp. The response time will be based on the last message with the same activityID.

mcharawi:
All comments addressed @charlierudolph

go/exocom/src/logger/logger.go
```diff
@@ -77,7 +78,7 @@ func (l *Logger) Messages(message structs.Message, originalName string, internal
 			text += fmt.Sprintf("--[ %s ]-[ %s ]->", originalName, internalMessageName)
 		}
 		text += "  " + receiver
-		if message.ResponseTo != "" {
+		if message.ResponseTime > time.Nanosecond {
```
charlierudolph: Is the default value 0? If so thoughts on comparing  `!= 0`? I think thats a bit more clear

mcharawi: The Duration type is technically just a wrapper for the int64 type, so the default is zero. Changed.

PR #188, opened 2017-08-01
fix random nil pointer dereference

### LINKED ISSUE #182: go exorelay: fix random nil pointer dereference
Ran into something like this a couple times
https://travis-ci.org/Originate/exocom/builds/258327182.

I believe this has to do with the fact that `e.socket` is set to nil in the `Close` method. This is done to prevent endless reconnects after calling `Close()`. Lets update to using a different internal variable to mark that it initialized the close and thus should not try to reconnect.

charlierudolph:
resolves #182


go/exorelay/exorelay.go
```diff
@@ -41,6 +42,7 @@ func (e *ExoRelay) Connect() error {
 	if err != nil {
 		return err
 	}
+	e.shouldReconnect = true
```
mcharawi: shouldnt it be `false` here and `true` on line 58?

PR #202, opened 2017-08-02
go libraries: replace responseTo in favor of activityId

mcharawi:
not too sure about the exoservice change


charlierudolph:
exoservice change looks good. exorelay needs to generate activity id if one is not provided.

mcharawi:
@charlierudolph comment addressed

go/exorelay/exorelay.go
```diff
@@ -65,27 +65,30 @@ func (e *ExoRelay) GetMessageChannel() chan structs.Message {
 }
 
 // Send sends a message with the given options
-func (e *ExoRelay) Send(options MessageOptions) (string, error) {
+func (e *ExoRelay) Send(options MessageOptions) (string, string, error) {
```
charlierudolph: I updated send on the javascript version to return the message instead of the just the message id. Thoughts on doing the same here?

mcharawi: Sounds good.

charlierudolph:
Please use descriptive PR titles instead of always using the default. For example, this PR should probably have the title `go: replace responseTo with activityId`. Please update the current PR and use this in the future.

mcharawi:
@charlierudolph comment addressed

go/exoservice/exoservice_test.go
```diff
@@ -72,9 +72,10 @@ func FeatureContext(s *godog.Suite) {
 
 	s.Step(`^receiving a "([^"]*)" message(?: with (?:id "([^"]*)")?(?:(?: and )?sessionId "([^"]*)")?)?$`, func(name, id, sessionId string) error {
 		message := structs.Message{
-			ID:        id,
-			Name:      name,
-			SessionID: sessionId,
+			ActivityID: id,
+			ID:         id,
```
charlierudolph: id isn't as important anymore and we should use different id and activity id. Please update the step to accept an activityId instead of an id and generate a random message id.

go/exorelay/exorelay.go
```diff
@@ -65,27 +65,30 @@ func (e *ExoRelay) GetMessageChannel() chan structs.Message {
...
+		return &structs.Message{}, errors.New("ExoRelay#Send not connected to Exocom")
 	}
 	id := uuid.NewV4().String()
 	if options.Name == "" {
-		return "", errors.New("ExoRelay#Send cannot send empty messages")
+		return &structs.Message{}, errors.New("ExoRelay#Send cannot send empty messages")
+	}
+	if options.ActivityID == "" {
+		options.ActivityID = uuid.NewV4().String()
 	}
```
charlierudolph: Thoughts on creating a local activityID variable instead of updating the passed in options? I prefer to not mutate inputs. Not a big deal in the current case as we are getting a copy of the messageOptions but if we ever changed it to a pointer then we would be updating the passed in options. This could be bad if the caller wanted to use it again after and wasn't expecting this function to change it.

go/websocketbridge/websocket_bridge_test.go
```diff
@@ -46,13 +46,15 @@ func FeatureContext(s *godog.Suite) {
...
 	var outgoingMessageId string
+	var outgoingActivityId string
 	var savedError error
 
 	s.BeforeScenario(func(interface{}) {
 		exocomPort = freeport.GetPort()
 		clientPort = freeport.GetPort()
 		clientURL = fmt.Sprintf("ws://localhost:%v", clientPort)
 		outgoingMessageId = ""
+		outgoingActivityId = ""
```
charlierudolph: This variable doesn't appear to be used after its assigned. Thoughts on renaming it to `expectedActivityId` and then using it on line 169?

mcharawi: We can't capture the value of the activity id since it's assigned in exorelay, but this can be removed.

mcharawi:

go/websocketbridge/websocket_bridge_test.go
```diff
@@ -163,6 +163,7 @@ func FeatureContext(s *godog.Suite) {
 		//ID and SessionId get generated on the fly, so we can't know them beforehand
```
charlierudolph: I believe this comment needs to be updated to include the activityId

PR #214, opened 2017-08-08
update message cache test

charlierudolph:
Still seeing some random failures around clearing the cache. Updated it to actually wait for it to clear the cache with a 10 second timeout (super safe)


mcharawi:
the safety of this is dependent on the go-routine scheduler, correct? i don't think we can ever guarantee a 100% success rate given the structure of the test, since there is always the likelihood that the `clearCache` go-routine wont be scheduled before this test runs.

charlierudolph:
Correct. This is just better in that we now have a timeout and are polling to check if its successful instead of having a dedicated wait time. Thus this test is more stable since waitFor has a timeout of 10 seconds but still fast since it finishes right after the cache gets cleared

PR #219, opened 2017-08-09
add exocom-lint

### LINKED ISSUE #207: add exocom-lint
Bring over exo-lint from exosphere. Defined [here](https://github.com/Originate/exosphere/tree/v0.22.1/exo-lint). Place in `javascript/exocom-lint`. Bring over any code needed that was from exosphere-shared.

charlierudolph:
resolves #207


javascript/exocom-lint/features/lint.feature
```diff
@@ -0,0 +1,40 @@
...
+
+  Rules:
+  - run "exo lint" in the directory of your application to lint it
+  - if a sent message is not received anywhere, it prints an error message
+  - if a received message it not sent anywhere, it prints an error message
+
+  Scenario: functioning application
+    Given I am in the directory of an application with the services:
+      | NAME  | SENDS                      | RECEIVES                   |
+      | web   | user.create, user.delete   | user.created, user.deleted |
```
mcharawi: Maybe use spaces instead of periods, i.e. `user create`?

javascript/exocom-lint/features/lint.feature
```diff
@@ -0,0 +1,40 @@
...
+  So that I know that all my messages are handled.
+
+  Rules:
+  - run "exo lint" in the directory of your application to lint it
+  - if a sent message is not received anywhere, it prints an error message
+  - if a received message it not sent anywhere, it prints an error message
+
+  Scenario: functioning application
+    Given I am in the directory of an application with the services:
+      | NAME  | SENDS                      | RECEIVES                   |
```
mcharawi: If these `sends` and `receives` values are coming from the service.yml, I believe that they list internal message names, not public, so it might be prudent to write a test for that case.

charlierudolph: Good catch. I'll add it an example that includes message translation

javascript/exocom-lint/features/lint.feature
charlierudolph: Will do

PR #234, opened 2017-08-16
add go race detector

charlierudolph:
resolves Originate/exosphere#462


charlierudolph:
Sounds good

charlierudolph:
@kevgo updated

go/exocom/test_helpers/mock_service.go
```diff
@@ -69,10 +71,9 @@ func (m *MockService) Close() error {
 
 //GetReceivedMessages gets the messages this mock service has received
 func (m *MockService) GetReceivedMessages() []structs.Message {
-	if m.ReceivedMessages == nil {
-		return []structs.Message{}
-	}
-	return m.ReceivedMessages
+	m.receivedMessagesMutex.RLock()
+	defer m.receivedMessagesMutex.RUnlock()
+	return m.receivedMessages
```
kevgo: Just out of curiosity: is the act of simply returning a variable already a race condition?

go/exoservice/exoservice.go
```diff
@@ -93,9 +95,11 @@ func (e *ExoService) ListenForMessages(messageHandlers MessageHandlerMapping) er
 			return nil
 		}
 		go e.receiveMessage(message)
+		e.replyChannelsMutex.RLock()
 		if replyChannel, ok := e.replyChannels[message.ActivityID]; ok {
 			replyChannel <- &message
 		}
+		e.replyChannelsMutex.RUnlock()
```
kevgo: Here you also send the message inside the read lock. Is that intentional? Or is the read lock only there to get the reply channel instance?

go/exocom/test_helpers/mock_service.go
charlierudolph: The race condition is around that property: `m.receivedMessages`. It is read here and written by a goroutine separately.

go/exoservice/exoservice.go
charlierudolph: The lock should only be around the read. Updated

go/frontendbridge/main.go
```diff
@@ -69,15 +71,15 @@ func (w *FrontendBridge) addConnection(auth structs.MessageAuth, socket *websock
...
+	w.connectionsMutex.RLock()
 	var result *Connection
 	for _, connection := range w.connections {
 		if reflect.DeepEqual(auth, connection.auth) {
 			result = connection
 			break
 		}
 	}
-	w.connectionsMutex.Unlock()
+	w.connectionsMutex.RUnlock()
```
kevgo: You are doing the entire loop inside the lock. I'm not 100% sure what the best way to use locks is, but I think I remember that you should keep the code path inside the lock as short as possible, since locks are global bottlenecks.

go/frontendbridge/main.go
```diff
@@ -156,21 +160,24 @@ func (w *FrontendBridge) Close() error {
...
 	w.addConnection(auth, socket)
 	utils.ListenForMessages(socket, func(message structs.Message) error {
+		w.exoRelayMutex.Lock()
 		_, err := w.exoRelay.Send(exorelay.MessageOptions{
 			Name:       message.Name,
 			Payload:    message.Payload,
 			ActivityID: message.ActivityID,
 			Auth:       auth,
 		})
+		w.exoRelayMutex.Unlock()
```
kevgo: What race condition does this write lock guard against?

go/exocom/test_helpers/mock_service.go
kevgo: Interesting, thanks for the explanation!

kevgo: I'd say if an update of this variable happens in parallel to reading it, returning an older value is not necessarily wrong behavior. Or can the value we read here be garbage while being updated somewhere else?

charlierudolph:
All these locks were added because go race detector detected data races.

charlierudolph:
The blog post on go race detector didn't suggest locks but that was the primary way I was thinking for fixing this. One idea from the blog post was using channels so only a single goroutine is in control of the variable.

go/frontendbridge/main.go
charlierudolph: I am searching through the available connections looking for one where the auth matches. This should be fast and I can't think of any better way to do this

go/frontendbridge/main.go
charlierudolph: the race detector alerted about a race condition in writing to the websocket of exorelay.

kevgo:
In general, Go Goroutines are the higher-level and more idiot-proof concept, locks only exist for certain performance-critical situations where we don't want to keep sending data back and forth. Having said that, the [Go Wiki](https://github.com/golang/go/wiki/MutexOrChannel) is pretty pragmatic about this. Good to ship from my end!

PR #236, opened 2017-08-18
Security Adapter Part One

mcharawi:
Gotta cleanup still

Addresses https://github.com/Originate/exosphere/issues/460


hugobho:
Linter seems to be failing though:
```
src/exocom/exocom.go:103::warning: cyclomatic complexity 11 of function (*ExoCom).listenForMessages() is high (> 10) (gocyclo)
```

go/exocom/src/exocom/exocom.go
```diff
@@ -95,16 +100,40 @@ func (e *ExoCom) listenForErrors() {
...
-		if e.routingManager.CanSend(message.Sender, message.Name) {
-			err := e.send(message)
-			if err != nil {
-				printLogError(e.logger.Error(err.Error()))
-			}
+		if !e.hasSecurity {
+			e.listenSender(message)
 		} else {
-			printLogError(e.logger.Warning(fmt.Sprintf("Warning: Service '%s' is not allowed to broadcast the message '%s'", message.Sender, message.Name)))
+			switch message.Name {
```
charlierudolph: Okay now that we have this logic working, can we try to extract it to a security manager struct? I imagine it could look something like:
```go
securityResult := e.securityManager.ReceiveMessage(message)
if securityResult.MessageToSend != nil {
  //...
} 
if securityResult.WarningMessageToLog != "" {
  //...
}
```
Where the securityManager could have a switch statement similar to the current one and encapsulates most of the logic that was added into exocom currently.

For the default case, the messageToSend is the serviceMessage created in the `sendToSecurity` method. It could then be passed to `e.send` since the security service is the only one that receives messages with the name "authorize message".

go/exocom/src/security/securty.go
```diff
@@ -0,0 +1,57 @@
...
+// NewSecurityManager returns a new security manager
+func NewSecurityManager() *Manager {
+	result := &Manager{
+		authTable: map[string]structs.Message{},
+	}
+	return result
+}
+
+// ReceiveMessage handles incoming security messages
+func (m *Manager) ReceiveMessage(message structs.Message) *Result {
```
charlierudolph: This method is now nicely unit testable. Thoughts on adding some unit tests for it? Mainly makes it easier to develop it (especially for part 2) as the unit tests will run much faster than the feature tests

go/exocom/src/exocom/exocom.go
```diff
@@ -98,13 +105,26 @@ func (e *ExoCom) listenForErrors() {
...
 			printLogError(e.logger.Warning(fmt.Sprintf("Warning: Service '%s' is not allowed to broadcast the message '%s'", message.Sender, message.Name)))
+		} else {
+			if !e.hasSecurity {
+				err := e.send(message)
+				if err != nil {
+					printLogError(e.logger.Error(err.Error()))
+				}
+			} else {
+				securityResult := e.securityManager.ReceiveMessage(message)
+				if securityResult.MessageToSend != (structs.Message{}) {
```
charlierudolph: Thoughts on making MessageToSend a pointer to a message so we can compare against nil here?

go/exocom/src/exocom/exocom.go
```diff
@@ -9,16 +9,20 @@ import (
...
+	"github.com/Originate/exocom/go/exocom/src/security"
 	"github.com/Originate/exocom/go/exocom/src/types"
 	"github.com/Originate/exocom/go/structs"
 	"github.com/gorilla/websocket"
 )
 
 // ExoCom is the top level message broadcaster
 type ExoCom struct {
+	securityManager   *security.Manager
+	authTable         map[string]structs.Message
```
charlierudolph: I believe this can be removed.

go/exocom/src/exocom/exocom.go
```diff
@@ -9,16 +9,20 @@ import (
...
 	connectionManager *connection.Manager
 	deregisterChannel chan string
 	errorChannel      chan error
+	hasSecurity       bool
```
charlierudolph: Thoughts on passing this through to the security manager? Then we don't need to store it here and the security manager will return the passed in message when this is false.

go/exocom/src/exocom/exocom.go
```diff
@@ -30,10 +34,13 @@ type ExoCom struct {
 var upgrader = websocket.Upgrader{}
 
 // New creates a new ExoCom instance
-func New(serviceRoutes types.Routes) (*ExoCom, error) {
+func New(serviceRoutes types.Routes, hasSecurity bool) (*ExoCom, error) {
```
charlierudolph: I don't think hasSecutiry needs to be passed in. We could call serviceRoutes.GetHasSecurity() in this function

go/exocom/src/security/security_test.go
```diff
@@ -2,13 +2,44 @@ package security_test
...
+			manager         *security.Manager
+			incomingMessage structs.Message
+			securityRequest *security.Result
+			securityResult  *security.Result
+		)
+
+		BeforeEach(func() {
+			manager = security.NewSecurityManager()
+			incomingMessage = structs.Message{
+				Name: "name",
```
charlierudolph: Can we give this a more realistic name like `create user`?

go/exocom/src/security/security_test.go
```diff
@@ -2,13 +2,44 @@ package security_test
...
+			Expect(securityResult.MessageToSend).To(Equal(incomingMessage))
+			Expect(securityResult.WarningMessage).To(Equal(""))
+		})
+		It("finds and returns authorized messages", func() {
+			message := structs.Message{
+				Name: "message unauthorized",
+			}
+			securityResult = manager.ReceiveMessage(message)
+			Expect(securityResult.MessageToSend).To(Equal(structs.Message{}))
+			Expect(securityResult.WarningMessage).To(Equal("Warning: Unauthorized message '' from '' with activityId ''"))
```
charlierudolph: Can we update this test case so the warning message is entirely filled out? I think we need to pass an activityId with "message unauthorized" and set a Sender on the initialMessage

go/exocom/src/security/security_test.go
```diff
@@ -2,13 +2,44 @@ package security_test
...
+				Name: "name",
+			}
+			securityRequest = manager.ReceiveMessage(incomingMessage)
+		})
+
+		It("finds and returns authorized messages", func() {
+			message := structs.Message{
+				Name:       "message authorized",
+				ActivityID: securityRequest.MessageToSend.ActivityID,
+			}
```
charlierudolph: Thoughts on removing this variable and defining the message inline with the ReceiveMessage call
```golang
securityResult = manager.ReceiveMessage(structs.Message{
  //...
})
```
Same for unauythorized

go/exocom/src/security/security_test.go
```diff
@@ -2,13 +2,44 @@ package security_test
...
 )
 
 var _ = Describe("Security", func() {
-	It("creates a security manager", func() {
-		securityManager := security.NewSecurityManager()
-		Expect(securityManager).ToNot(BeNil())
+	Describe("ReceiveMessage", func() {
+		var (
+			manager         *security.Manager
+			incomingMessage structs.Message
```
charlierudolph: Thoughts on calling this initialMessage?

go/exocom/src/security/security_test.go
```diff
@@ -2,13 +2,44 @@ package security_test
...
+			securityRequest *security.Result
```
charlierudolph: Thoughts on naming this `initialSecurityResult`

go/exocom/src/security/security_test.go
```diff
@@ -2,13 +2,44 @@ package security_test
...
+			securityResult  *security.Result
```
charlierudolph: I don't think this var is needed since securityResult is not shared across code blocks

go/exocom/src/security/security_test.go
```diff
@@ -2,13 +2,44 @@ package security_test
...
+		It("finds and returns authorized messages", func() {
+			message := structs.Message{
+				Name:       "message authorized",
+				ActivityID: securityRequest.MessageToSend.ActivityID,
+			}
+			securityResult = manager.ReceiveMessage(message)
+			Expect(securityResult.MessageToSend).To(Equal(incomingMessage))
+			Expect(securityResult.WarningMessage).To(Equal(""))
+		})
+		It("finds and returns authorized messages", func() {
```
charlierudolph: I believe this test description needs to be updated: authorized -> unauthorized

go/exocom/src/security/securty.go
mcharawi: You bet

PR #238, opened 2017-08-25
Security Adapter Part 2

mcharawi:
Still a bit messy, want to get some eyes before the weekend


charlierudolph:
The changes to the dependencies don't appear to be used or needed. Can you revert the changes to glide.lock, glide.yaml, and the vendor folder.

go/exocom/src/security/manager.go
```diff
@@ -33,6 +34,17 @@ func (m *Manager) ReceiveMessage(message structs.Message) *Result {
...
+		bytes, err := json.Marshal(message.Payload)
+		if err != nil {
+			warningMessage = err.Error()
+		}
+		messageToSend = &structs.Message{}
+		err = json.Unmarshal(bytes, messageToSend)
+		if err != nil {
+			warningMessage = err.Error()
+		}
+		m.authTable[message.ActivityID] = &structs.Message{}
```
charlierudolph: Can we use a separate table for security requests? I don't like overloading the authTable. Also `message.ActivityID` should already contain something. The flow would be something like this:

* exocom receives "create user" that needs to be authorized
  * exocom sends "authorize message" to security service with activityId 123
* execom receives "security request" with activityId 123
  * execom sends the message in that payload (assume the message in the payload has activityId 456)
* exocom receives message with activityId 456 and isSecurity flag 
  * exocom sends "security response" with activityId 123 and the received message in the payload 
* exocom receives "message authorized" or "message unauthorized" with activityId 123
  * ...

go/structs/message.go
```diff
@@ -20,3 +25,27 @@ type MessagePayload interface{}
...
+
+func (s *Message) FillMessage(m map[string]interface{}) error {
+	for k, v := range m {
+		err := setField(s, strings.Title(k), v)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
charlierudolph: I believe these methods can be removed

go/exocom/features/security-adapter.feature
```diff
@@ -100,12 +100,101 @@ Feature: Security Adapter
...
+  Scenario: security request
+    When the "security" service sends:
+      """
+      {
+        "name": "security request",
+        "payload": {
+          "name": "get session",
+          "payload": {
+            "id": "2"
+          }
```
charlierudolph: Lets assume this has an id and activityId in the payload as well.

go/exocom/main_test.go
```diff
@@ -189,16 +189,19 @@ func FeatureContext(s *godog.Suite) {
...
+
+		expectedBytes, err := json.Marshal(expectedMessage.Payload)
+		actualBytes, err := json.Marshal(actualMessage.Payload)
+		actualMessagePayload := structs.Message{}
+		expectedMessagePayload := structs.Message{}
+		err = json.Unmarshal(actualBytes, &actualMessagePayload)
+		err = json.Unmarshal(expectedBytes, &expectedMessagePayload)
+		actualMessagePayload.Sender = expectedMessagePayload.Sender
+		actualMessagePayload.ID = expectedMessagePayload.ID
+		actualMessagePayload.IsSecurity = expectedMessagePayload.IsSecurity
```
charlierudolph: This shouldn't need to be assigned as we should be able to determine it

mcharawi:
@charlierudolph all comments addressed

go/exocom/main_test.go
```diff
@@ -149,7 +149,7 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`^the "([^"]+)" service sends "([^"]*)"$`, func(serviceName, messageName string) error {
 		message := structs.Message{
 			ID:      "123",
-			Payload: "",
+			Payload: map[string]interface{}{},
```
charlierudolph: Can this be reverted?

go/exocom/main_test.go
```diff
@@ -189,16 +189,18 @@ func FeatureContext(s *godog.Suite) {
...
-		}
-		delete(payload, "sessionId")
-		delete(payload, "responseTime")
-		delete(payload, "sender")
-		delete(payload, "timestamp")
-		if !reflect.DeepEqual(expectedMessage.Payload, payload) {
-			return fmt.Errorf("Expected payload to equal %s but got %s", expectedMessage.Payload, actualMessage.Payload)
+
+		expectedBytes, err := json.Marshal(expectedMessage.Payload)
+		actualBytes, err := json.Marshal(actualMessage.Payload)
```
charlierudolph: There are some unchecked errors here

go/exocom/main_test.go
```diff
@@ -251,7 +253,7 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`^the "([^"]*)" service sends "([^"]*)" for activity "([^"]*)"$`, func(serviceName, replyMessage, activityId string) error {
 		return services[serviceName].Send(structs.Message{
 			Sender:     serviceName,
-			Payload:    "",
+			Payload:    map[string]interface{}{},
```
charlierudolph: Can this be reverted?

go/exocom/main_test.go
```diff
@@ -189,16 +189,18 @@ func FeatureContext(s *godog.Suite) {
...
-		if !reflect.DeepEqual(expectedMessage.Payload, payload) {
-			return fmt.Errorf("Expected payload to equal %s but got %s", expectedMessage.Payload, actualMessage.Payload)
+
+		expectedBytes, err := json.Marshal(expectedMessage.Payload)
+		actualBytes, err := json.Marshal(actualMessage.Payload)
+		actualMessagePayload := structs.Message{}
+		expectedMessagePayload := structs.Message{}
+		err = json.Unmarshal(actualBytes, &actualMessagePayload)
+		err = json.Unmarshal(expectedBytes, &expectedMessagePayload)
+		actualMessagePayload.Sender = expectedMessagePayload.Sender
```
charlierudolph: Why is this being overwritten? The sender seems like something we should be able to hardcode in the JSON of the feature

go/exocom/src/security/manager.go
```diff
@@ -33,6 +36,17 @@ func (m *Manager) ReceiveMessage(message structs.Message) *Result {
 	var messageToSend *structs.Message
 	warningMessage := ""
 	switch message.Name {
+	case "security request":
+		bytes, err := json.Marshal(message.Payload)
+		if err != nil {
+			warningMessage = err.Error()
```
charlierudolph: Please wrap these errors with https://github.com/pkg/errors to give more context. For example here
`warningMessage = errors.Wrap(err, "Marshaling 'security request' payload").Error()`

go/exocom/main_test.go
```diff
@@ -189,16 +189,29 @@ func FeatureContext(s *godog.Suite) {
 			return fmt.Errorf("Expected to receive a message but got %v", err)
 		}
 		outgoingMessageActivityID = actualMessage.ActivityID
-		payload, ok := actualMessage.Payload.(map[string]interface{})
-		if !ok {
-			return errors.New("Error: cannot parse payload")
+
+		expectedBytes, err := json.Marshal(expectedMessage.Payload)
+		if err != nil {
+			return fmt.Errorf(err.Error())
```
charlierudolph: This can be `return err`. Same for the other instances in this function

go/exocom/main_test.go
```diff
@@ -189,16 +189,29 @@ func FeatureContext(s *godog.Suite) {
...
 		}
-		delete(payload, "sessionId")
-		delete(payload, "responseTime")
-		delete(payload, "sender")
-		delete(payload, "timestamp")
-		if !reflect.DeepEqual(expectedMessage.Payload, payload) {
-			return fmt.Errorf("Expected payload to equal %s but got %s", expectedMessage.Payload, actualMessage.Payload)
+		actualMessagePayload.ID = expectedMessagePayload.ID
+		actualMessagePayload.ActivityID = expectedMessagePayload.ActivityID
```
charlierudolph: Are these two line even needed? Based on the added tests, they don't appear to be as the id and activityId are not generated by exocom but passed in

go/exocom/src/security/manager.go
```diff
@@ -33,6 +37,17 @@ func (m *Manager) ReceiveMessage(message structs.Message) *Result {
 	var messageToSend *structs.Message
 	warningMessage := ""
 	switch message.Name {
+	case "security request":
+		bytes, err := json.Marshal(message.Payload)
+		if err != nil {
+			warningMessage = errors.Wrap(err, "Marshaling 'security request' payload").Error()
```
charlierudolph: I believe we need a `break` after this to jump to the end of the switch statement and prevent processing any further. Same for the other instance we set warningMessage

go/exocom/main_test.go
```diff
@@ -189,16 +189,27 @@ func FeatureContext(s *godog.Suite) {
...
+		}
+		actualBytes, err := json.Marshal(actualMessage.Payload)
+		if err != nil {
+			return err
+		}
+		actualMessagePayload := structs.Message{}
+		expectedMessagePayload := structs.Message{}
+		err = json.Unmarshal(actualBytes, &actualMessagePayload)
+		if err != nil {
+			return fmt.Errorf(err.Error())
```
charlierudolph: This can be simplified to `return err` as well. Also line 209.

PR #240, opened 2017-08-29
Security Adapter Functionality for Exorelay/Exoservice

mcharawi:


charlierudolph:
We also need this for the javascript implementations. Can we add that to this PR as well?

mcharawi:
@charlierudolph added

javascript/exorelay/features/outgoing-replies.feature
```diff
@@ -114,3 +114,33 @@ Feature: Sending outgoing replies to incoming messages
       activity-id: '456'
       auth: '1'
       """
+
+
+  Scenario: sending a reply with security flag
```
charlierudolph: This sounds like the reply is setting a flag whereas the flag is on the incoming message. thoughts on updating this to "sending a reply when the incoming message has the security flag"

mcharawi: 👍

PR #241, opened 2017-08-30
javascript/exoservice: update interface

charlierudolph:
See the new test fixtures. Require exoservice and pass the handlers into a bootstrap method.

This allows uses to write the handlers in any compile to JS language and they can run the file however they want. This is very similar to the go API


javascript/exoservice/features/incoming-requests.feature
```diff
@@ -15,7 +15,7 @@ Feature: Handling incoming messages
 
   Background:
     Given an ExoCom instance
-    And an instance of the "test service" service
+    And an instance of the "test-service" fixture
```
kevgo: `fixture` is a much better name!

PR #245, opened 2017-09-05
Exocom: use default port 80

hugobho:
resolves https://github.com/Originate/exosphere/issues/514


javascript/exorelay/src/exorelay.ls
```diff
@@ -11,7 +11,7 @@ class ExoRelay extends EventEmitter
 
   (@config) ->
     @config?.exocom-host or throw new Error 'ExoCom host not provided to Exorelay'
-    @config?.exocom-port or throw new Error 'ExoCom port not provided to Exorelay'
+    @config.exocom-port? or @config.exocom-port = 80
```
charlierudolph: Thoughts on update this function to start with
```
(@config = {}) ->
```
Then we can remove the `?` from the exocom-host and role as well.

charlierudolph: I believe we can simplify this line to  `@config.exocom-port or= 80`

go/exoservice/exoservice.go
```diff
@@ -32,17 +32,14 @@ type ReplyChannelMapping map[string]chan *structs.Message
 // Bootstrap brings an ExoService online using environment variables and the
 // given messageHandlers
 func Bootstrap(messageHandlers MessageHandlerMapping) {
-	port, err := strconv.Atoi(os.Getenv("EXOCOM_PORT"))
-	if err != nil {
-		panic(fmt.Sprintf("Invalid port: %s", os.Getenv("EXOCOM_PORT")))
-	}
+	port := getPort()
```
charlierudolph: Can this variable be removed and we can `getPort()` in the definition of `config`

hugobho: I did that at first but there's a print statement on line 46 that needs this var

charlierudolph: Danget, I looked for that and couldn't find it. Github did not expand the diff the way I expected. Good as it

hugobho:

PR #248, opened 2017-09-15
Exosecurity-JavaScript

mcharawi:
@charlierudolph


javascript/exosecurity/src/exosecurity.js
```diff
@@ -0,0 +1,58 @@
...
+    })
+    this.handlers = {
+      'authorize message': (message, { reply }, activityId) =>
+        messageAuthorizer(message, this.buildRequester(activityId), result => {
+          if (result) {
+            return reply('message authorized')
+          }
+          return reply('message unauthorized')
+        }),
+    }
```
charlierudolph: Thoughts on removing `this.handlers` and defining the handlers in in the connect method as that is the only place they are used

javascript/exosecurity/src/exosecurity.js
```diff
@@ -0,0 +1,58 @@
...
+      exocomPort,
+      role,
+    })
+    this.handlers = {
+      'authorize message': (message, { reply }, activityId) =>
+        messageAuthorizer(message, this.buildRequester(activityId), result => {
+          if (result) {
+            return reply('message authorized')
+          }
+          return reply('message unauthorized')
```
charlierudolph: thoughts on drying this up by defining a messageName variable? Then we would only have one use of `reply`

javascript/exosecurity/src/exosecurity.js
```diff
@@ -0,0 +1,58 @@
...
+      from: this.exoRelay,
+      to: this,
+    })
+  }
+
+  buildRequester(activityId) {
+    return (name, payload, callback) => {
+      const request = { name, payload }
+      request.activityId = uuid.v1()
+      request.id = uuid.v1()
```
charlierudolph: These three lines could be a single statement
```js
const request = {name, payload, activityId: uuid.v1(), id: uuid.v1()}
```

javascript/exosecurity/features/support/steps-then.js
```diff
@@ -0,0 +1,17 @@
+const jsDiff = require('jsdiff-console')
```
charlierudolph: Please remove this in favor of [chai](http://chaijs.com/)

javascript/exosecurity/features/support/env.js
```diff
@@ -0,0 +1,30 @@
...
+const World = require('./world')
+
+setDefaultTimeout(1000)
+setWorldConstructor(World)
+
+function closeIfDefined(obj, done) {
+  if (obj) {
+    obj.close(done)
+  }
+  done()
```
charlierudolph: This looks wrong to me. I believe this needs an `else` to prevent a fall through.

javascript/exosecurity/features/support/steps-given.js
```diff
@@ -0,0 +1,35 @@
+const { defineSupportCode } = require('cucumber')
```
charlierudolph: `Given` can be required directly from cucumber. Same with When / Then. Lets use that instead of `defineSupportCode`

javascript/exosecurity/features/support/steps-given.js
```diff
@@ -0,0 +1,35 @@
...
+        this.exocomPort = exocomPort
+        this.exocom = new ExoComMock()
+        this.exocom.listen(this.exocomPort)
+        done()
+      })
+    )
+  })
+  steps.Given(/^receiving this message:$/, function(message) {
+    message = JSON.parse(message)
+    message.service = this.role // eslint-disable-line no-param-reassign
```
charlierudolph: I think can avoid disabling this rule if we make the input parameter `messageStr` (since its a string anyway) and then message is a local variable which cause this rule to no longer apply.

javascript/exosecurity/features/support/world.js
```diff
@@ -0,0 +1,46 @@
...
+        ROLE: role,
+      },
+      verbose: observableProcessOptions.verbose,
+      stdout: observableProcessOptions.stdout,
+      stderr: observableProcessOptions.stderr,
+    })
+    this.process.wait('online at port', done)
+  }
+  this.removeRegisterServiceMessage = function(exocom, done) {
+    this.exocom = exocom
```
charlierudolph: I don't think this assignment is needed

javascript/exosecurity/features/support/world.js
```diff
@@ -0,0 +1,46 @@
...
+    return waitUntil(() => this.exocom.receivedMessages.length, 10, () => {
```
charlierudolph: The return is unneeded as what the function returns is unimportant. It is passed a callback instead

mcharawi:
@charlierudolph all comments addressed

javascript/exosecurity/features/support/world.js
```diff
@@ -0,0 +1,45 @@
+const ObservableProcess = require('observable-process')
+const path = require('path')
+const { any } = require('prelude-ls')
```
charlierudolph: Can we remove `prelude-ls`? Feels weird to use that now that we aren't using livescript.  We can use lodash in order to have an `any` function

PR #250, opened 2017-09-19
Updating Exorelay.Connect retry to have exponential backoff

mcharawi:
addresses https://github.com/Originate/exosphere/issues/572


go/utils/retry.go
```diff
@@ -4,11 +4,13 @@ import "time"
 
 // Retry will execute the given function until it returns true with the given
 // delay (milliseconds) between calls
```
charlierudolph: This comment needs to be updated

javascript/exorelay/src/websocket-connector/websocket-connector.ls
```diff
@@ -71,6 +71,13 @@ class WebSocketConnector extends EventEmitter
       ..on 'message', @_on-socket-message
       ..on 'error', @_on-socket-error
       ..on 'close', @_on-socket-close
+    @retry 100
```
charlierudolph: I don't think this prevents the swamp of error messages that gets logged as the library continually emits error messages which we log. We should only log those unable to connect errors with exponential backoff

go/utils/websocket.go
```diff
@@ -9,10 +9,10 @@ import (
 // ConnectWithRetry will attempt to connect a websocket to the given url
 // retrying if needed with the given delay (milliseconds) between calls
 func ConnectWithRetry(url string, delay int) (socket *websocket.Conn, err error) {
-	Retry(delay, func() bool {
+	Retry(delay, func(updatedDelay int) bool {
 		socket, _, err = websocket.DefaultDialer.Dial(url, nil)
 		if err != nil {
-			fmt.Printf("Unable to connect (%s). Retrying...\n", err)
+			fmt.Printf("Unable to connect (%s). Retrying after %v ms...\n", err, updatedDelay*2)
```
charlierudolph: I don't believe this log is accurate, as this fn is passed the initialDelay and the not the nextDelay

go/utils/websocket.go
```diff
@@ -9,10 +9,10 @@ import (
 // ConnectWithRetry will attempt to connect a websocket to the given url
 // retrying if needed with the given delay (milliseconds) between calls
```
charlierudolph: I believe this comment needs to be updated as well

go/utils/websocket.go
mcharawi: yes, but the `initialDelay` is updated in the `Retry` method's `for` loop, the log is anticipating the doubling of the delay, hence the `updatedDelay * 2`

go/utils/retry.go
```diff
@@ -4,11 +4,13 @@ import "time"
...
-func Retry(delayInMilliseconds int, attemptFn func() bool) {
+func Retry(delayInMilliseconds int, attemptFn func(int) bool) {
+	initialDelay := delayInMilliseconds
 	for {
-		if attemptFn() {
+		if attemptFn(initialDelay) {
 			return
 		}
-		time.Sleep(time.Duration(delayInMilliseconds) * time.Millisecond)
+		initialDelay = initialDelay * 2
```
charlierudolph: Can we move this multiplication to after the sleep so that? Currently we never sleep the passed in amount (the first amount that is slept if double that value). This also sends in the next delay to the attemptFn so it won't need to double the passed in value

charlierudolph:
Closing as deciding to go another way. All I care about is not printing the log messages every second so going to deal with that in another way

PR #253, opened 2017-09-22
Purge activity Ids

hugobho:
@charlierudolph  any suggestions for testing this?


go/exocom/src/connection/manager.go
```diff
@@ -17,18 +18,29 @@ type Manager struct {
...
+		activityIDCache:             map[string]time.Time{},
 	}
+	go func() {
+		for {
+			time.Sleep(options.ActivityIDCleanupInterval)
+			manager.servicesMutex.Lock()
+			manager.clearCache(options.ActivityIDCleanupInterval)
+			manager.servicesMutex.Unlock()
+		}
+	}()
```
charlierudolph: Thoughts on moving the func into a helper function so this becomes
```go
go manager.startActivityIDCleanup()
```

go/exocom/src/connection/nested_service_mapping.go
```diff
@@ -42,6 +42,20 @@ func (n NestedServiceMapping) DeleteForService(key1 string, service *Service) {
 	}
 }
 
+// DeleteForActivityID deletes entries of a given service that contain activityID
+func (n NestedServiceMapping) DeleteForActivityID(activityID string) {
```
charlierudolph: This struct doesn't know its holder activityIds. Can we make this more generic? Something like `DeleteNestedKey`. We could also add some unit tests for this function since we don't have a specific feature test for it

hugobho:

go/exocom/src/connection/manager.go
```diff
@@ -103,3 +110,25 @@ func (m *Manager) getServiceToSendMessageTo(role string, message structs.Message
 	}
 	return service
 }
+
+func (m *Manager) startActivityIDCleanup(cleanupInterval time.Duration) {
+	for {
+		time.Sleep(cleanupInterval)
+		m.servicesMutex.Lock()
+		m.clearCache(cleanupInterval)
+		m.servicesMutex.Unlock()
```
charlierudolph: Thoughts on moving the lock and unlock into the clearCache function?

go/exocom/src/connection/manager.go
```diff
@@ -17,18 +18,22 @@ type Manager struct {
...
+	manager := &Manager{
 		deregisterChannel:           options.DeregisterChannel,
 		errorChannel:                options.ErrorChannel,
 		messageChannel:              options.MessageChannel,
 		registerChannel:             options.RegisterChannel,
 		servicesByRoleAndID:         NestedServiceMapping{},
 		servicesByRoleAndActivityID: NestedServiceMapping{},
+		activityIDCache:             map[string]time.Time{},
 	}
+	manager.startActivityIDCleanup(options.ActivityIDCleanupInterval)
```
charlierudolph: I believe this needs to have `go` in front so it is done in its own goroutine.

Issue #267, opened 2017-09-28
exorelay: make responsibilities consistent


charlierudolph:
Currently the go and javascript forms are very different in what they are responsible for / what is left in exoservice. I believe this was a result of the very different times they were written and 

* exorelay
  * go
    * interface to receive all message
    * interface to send a message
  * js
    * interface to register handlers for specific messages
    * interface to send a message and receive the first reply
* exoservice
  * go 
    * interface to register handlers for specific message
  * js
    * small wrapper around exorelay (very little code)

---

I think we should make this consistent so that exorelay does the following things
* interface to receive all message
* interface to send a message

Thus exorelay is left as the low level way of dealing with exocom for anyone that wants to build on top of it instead of using exoservice. This was the idea we went with in the go version

---

Then we can update exoservice with a few methods:

* interface to send a message and receive the first reply 
  * currently implemented in js exorelay
  * the helper function used in go exoservice does this but it is not available for use standalone
* interface to register handlers for specific message 
  * currently implemented in js exorelay
  * currently implemented in go exoservice

---

If accepted, this would involve moving js code from exorelay to exoservice and adding a missing features to go exoservice.@kevgo @alexdavid @hugobho thoughts?

kevgo:
I am not against this. Just want to point out that the APIs for both ExoRelay-JS and Exoservice-JS were designed for two very specific use cases:
- exorelay: for front-end micro-services. In those situations one usually has an existing code base (for example an express-based web server), all we want to do is make it able to talk to Exocom. Typically I want to be able to handle particular messages. Not much need to also implement routing of messages.
- exoservice: for back-end micro-services. All I want here is handle incoming messages and send replies.

What is the use-case for the lower-level Exorelay where I get all received messages and then have to do the routing myself? 

If we find one, maybe we do a three-way split? (1) low-level Exorelay, (2) high-level Exorelay (1 + message router + reply handling), (3) Exoservice.

charlierudolph:
I'm suggesting moving high level exorelay into exoservice so it can be used for both (bootstrap for the main services and the ability to create an instance directly for the other methods)

I don't know the low level use case. In the go version, I believe it was in order to abstract away what we always want the exocom libraries to be responsible for (websockets / marshalling messages) and thats it. I believe that was mostly @alexdavid's idea where everything else was pushed out exoservice. I'm not certain if we want

kevgo:
[...]

PR #272, opened 2017-10-22
go: add exosocket

charlierudolph:
Moving `go/exorelay` to `go/exosocket`. This will be the low-level API while exorelay will become the mid-level api. This is mostly moving code. I'll call out the new code that was added

Part of #267


go/exosocket/config.go
```diff
@@ -0,0 +1,38 @@
...
+
+// Config contains the configuration values for ExoSocket instances
+type Config struct {
+	Host string
+	Port int
+	Role string
+}
+
+// NewEnvConfig returns a Config object based on environment variables
+func NewEnvConfig() Config {
```
charlierudolph: This is a new function. Will be used by exoservice and in tests.

go/exosocket/index.go
```diff
@@ -0,0 +1,106 @@
...
+type ExoSocket struct {
+	config                      Config
+	shouldReconnectOnDisconnect bool
+	socket                      *websocket.Conn
+	messageChannel              chan structs.Message
+}
+
+// NewExoSocket returns an ExoSocket with the given config
+func NewExoSocket(config Config) *ExoSocket {
+	return &ExoSocket{config: config}
```
charlierudolph: This is a new function as the object should not be instantiated directly

charlierudolph: Sounds good. I'll look into that. The interface would probably just include the public methods as the private fields are not used outside this file.

go/exosocket/index.go
```diff
@@ -0,0 +1,106 @@
...
+	go e.listenForMessages()
+	_, err = e.Send(MessageOptions{
+		Name: "exocom.register-service",
+	})
+	return err
+}
+
+// Close takes this ExoRelay instance offline
+func (e *ExoSocket) Close() error {
+	e.shouldReconnectOnDisconnect = false
```
Bowbaq: This looks like it may be called concurrently. If so, this needs to be protected by a mutex or manipulated using the `atomic` package to avoid a data race

go/exosocket/index.go
```diff
@@ -0,0 +1,106 @@
...
+	return e.messageChannel
+}
+
+// Send sends a message with the given options
+func (e *ExoSocket) Send(options MessageOptions) (*structs.Message, error) {
+	var activityID string
+	if e.socket == nil {
+		return &structs.Message{}, errors.New("ExoRelay#Send not connected to Exocom")
+	}
+	id := uuid.NewV4().String()
```
Bowbaq: No need to generate this here if you're just going to error out in the next step. I'd suggest putting this inline in the struct initializer

go/exosocket/index.go
```diff
@@ -0,0 +1,106 @@
...
+	if options.Name == "" {
+		return &structs.Message{}, errors.New("ExoRelay#Send cannot send empty messages")
+	}
+	if options.ActivityID == "" {
```
Bowbaq:
```go
activityID := options.ActivityID
if activityID == "" {
  activityID = uuid.NewV4().String()
}
```

go/exosocket/index.go
```diff
@@ -0,0 +1,106 @@
...
+func NewExoSocket(config Config) *ExoSocket {
+	return &ExoSocket{config: config}
+}
+
+// Connect brings an ExoSocket instance online
+func (e *ExoSocket) Connect() error {
+	if e.config.Port == 0 {
+		e.config.Port = 80
+	}
+	exocomURL := fmt.Sprintf("ws://%s:%d/services", e.config.Host, e.config.Port)
```
Bowbaq: What if `Host` is empty here? Consider using a`url.URL{Scheme: "ws", Host: ...}`.

go/exosocket/index.go
```diff
@@ -0,0 +1,106 @@
...
+	messageChannel              chan structs.Message
+}
+
+// NewExoSocket returns an ExoSocket with the given config
+func NewExoSocket(config Config) *ExoSocket {
+	return &ExoSocket{config: config}
+}
+
+// Connect brings an ExoSocket instance online
+func (e *ExoSocket) Connect() error {
```
Bowbaq: Any reads / writes in this method need to be synchronized since it is called in a goroutine by `listenForMessages`, otherwise there'll be data races

go/exosocket/index.go
```diff
@@ -0,0 +1,106 @@
...
+		return &structs.Message{}, err
+	}
+	return message, e.socket.WriteMessage(websocket.TextMessage, serializedBytes)
+}
+
+// Helpers
+
+func (e *ExoSocket) listenForMessages() {
+	utils.ListenForMessages(e.socket, func(message structs.Message) error {
+		e.messageChannel <- message
```
Bowbaq: I'm not entirely sure this works. The underlying channel can be replaced by `Connect` when reconnecting. It's unclear if the value captured by this closure would get updated as well

go/exosocket/test-fixtures/receiving_messages.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/Originate/exocom/go/exosocket"
+	"github.com/Originate/exocom/go/structs"
+	"github.com/Originate/exocom/go/utils"
+)
+
+// ReceivingMessagesTestFixture is a test fixture which saves the messages it receives
+type ReceivingMessagesTestFixture struct {
+	receivedMessages      []structs.Message
+	receivedMessagesMutex sync.RWMutex
```
Bowbaq: mutexes are placed above the data they protect in the struct by convention

go/exosocket/index_test.go
```diff
@@ -0,0 +1,279 @@
...
+		return err
+	})
+
+	s.Step(`^ExoCom goes offline momentarily`, func() error {
+		return exoComMockInstance.CloseConnection()
+	})
+}
+
+func TestMain(m *testing.M) {
+	var paths []string
```
Bowbaq: Use a single var keyword when declaring multiple vars next to each other
```go
var (
  paths []string
  format string
)
```

kevgo:
Thanks @Bowbaq!

go/exosocket/config.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+}
+
+func getPort() int {
+	defaultPort := 80
+	userPort := os.Getenv("EXOCOM_PORT")
+	if userPort == "" {
+		return defaultPort
+	}
+	port, err := strconv.Atoi(userPort)
```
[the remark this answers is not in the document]
charlierudolph: removed this parsing and made the port a string

go/exosocket/index.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// NewExoSocket returns an ExoSocket with the given config
+func NewExoSocket(config Config) *ExoSocket {
+	return &ExoSocket{config: config}
+}
+
+// Connect brings an ExoSocket instance online
+func (e *ExoSocket) Connect() error {
+	if e.config.Port == 0 {
```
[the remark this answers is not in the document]
charlierudolph: removed

go/exosocket/index.go
```diff
@@ -0,0 +1,106 @@
...
+		e.config.Port = 80
+	}
+	exocomURL := fmt.Sprintf("ws://%s:%d/services", e.config.Host, e.config.Port)
+	socket, err := utils.ConnectWithRetry(exocomURL, 100)
+	if err != nil {
+		return err
+	}
+	e.shouldReconnectOnDisconnect = true
+	e.socket = socket
+	e.messageChannel = make(chan structs.Message)
```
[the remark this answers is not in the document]
charlierudolph: How do you choose a buffer size? I have all the buffers being size one to synchronize them and feels like it has been simpler to debug since reads and write block. Makes the receiver do any longer processing in a separate goroutine.

go/exosocket/index.go
charlierudolph: updated. Don't think that changed much though as the Host and port are still passed in as a single string

go/exosocket/index.go
charlierudolph: Great catch. Moved the channel creation into the constructor so it is no longer re-created on each connect

go/exosocket/index.go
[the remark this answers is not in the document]
Bowbaq: Buffer size would depend on the expected throughput of the system, timeouts, memory constraints etc.

The default size of a channel is 0 (`make(chan structs.Message)` is equivalent to `make(chan structs.Message, 0)`). As per the [spec](https://golang.org/ref/spec#Send_statements) this means a receiver must be ready for any send to proceed.

It's probably a good idea to have a few second's worth of buffering here

PR #273, opened 2017-10-22
go: update frontendbridge to use exosocket

charlierudolph:
part of #267


go/frontendbridge/index.go
```diff
@@ -18,46 +18,48 @@ import (
...
+// NewFrontendBridge returns a FrontendBridge instance
+func NewFrontendBridge(config exosocket.Config, clientPort int) *FrontendBridge {
+	w := &FrontendBridge{
+		exoSocket:   exosocket.NewExoSocket(config),
+		connections: []*Connection{},
 	}
-	err = w.listenForClients(clientPort)
-	if err != nil {
-		return err
+	var handler http.HandlerFunc = func(resWriter http.ResponseWriter, req *http.Request) {
```
kevgo: Thoughts on extracting a `WebsocketServer` class here, which encapsulates the `http.Server` class? Then the FrontendBridge doesn't need to be concerned with upgrading basic HTTP connections to websockets, and that functionality becomes reusable.

PR #274, opened 2017-10-22
go: update exorelay

charlierudolph:
This also updates exoservice and exosecurity to used the new exorelay. Mostly moving code from exoservice to exorelay and thus a lot of simplification in exoservice and exosecurity

Part of #267


go/exorelay/index.go
```diff
@@ -0,0 +1,143 @@
...
+// RegisterHandler adds a handler for the given message name
+func (e *ExoRelay) RegisterHandler(messageName string, handler MessageHandler) {
+	e.messageHandlersMutex.Lock()
+	e.messageHandlers[messageName] = handler
+	e.messageHandlersMutex.Unlock()
+}
+
+// Send sends a message
+func (e *ExoRelay) Send(messageOptions exosocket.MessageOptions) (*structs.Message, error) {
+	return e.exoSocket.Send(messageOptions)
```
kevgo: Is this method necessary? Go provides infrastructure for composition already.

go/exorelay/index.go
```diff
@@ -0,0 +1,143 @@
...
+	case <-time.After(timeout):
+		return nil, fmt.Errorf("Did not receive a reply for '%s' after %s", activityID, timeout.String())
+	}
+}
+
+// Helpers
+
+func (e *ExoRelay) addReplyChannel(activityID string, messageChannel chan *structs.Message) {
+	e.replyChannelsMutex.Lock()
+	e.replyChannels[activityID] = messageChannel
```
kevgo: Can this accidentally override an already registered reply channel?

go/exorelay/index.go
```diff
@@ -0,0 +1,143 @@
...
+
+func (e *ExoRelay) removeReplyChannel(activityID string) {
+	e.replyChannelsMutex.Lock()
+	delete(e.replyChannels, activityID)
+	e.replyChannelsMutex.Unlock()
+}
+
+func (e *ExoRelay) sendToMessageHandler(message structs.Message) {
+	e.messageHandlersMutex.RLock()
+	defer e.messageHandlersMutex.RUnlock()
```
kevgo: This keeps the `messageHandlersMutex` open for the entirety of building the request, thereby limiting concurrency. Could you unlock the mutex already at line 131?

go/exorelay/index.go
```diff
@@ -0,0 +1,143 @@
...
+	handler, ok := e.messageHandlers[message.Name]
+	if ok {
+		handler(e.buildRequest(message))
+	}
+}
+
+func (e *ExoRelay) sendToReplyChannel(message structs.Message) {
+	e.replyChannelsMutex.RLock()
+	defer e.replyChannelsMutex.RUnlock()
```
kevgo: Same here. A best practice with locks is to unlock them as fast as possible.

go/exorelay/test-fixtures/register_handler/main.go
```diff
@@ -0,0 +1,39 @@
+package main
+
```
kevgo: I would add a little comment to these test fixtures explaining what their function is.

go/exosecurity/features/authorization.feature
```diff
@@ -10,7 +10,7 @@ Feature: Outgoing replies
 
 
   Background:
-    Given I connect the "simple" test fixture
+    Given I run the "simple" test fixture
```
kevgo: Less egocentric steps? `Given the "simple" test fixture runs`

go/exoservice/index.go
```diff
@@ -0,0 +1,27 @@
...
+// Bootstrap is the high level API to Exocom
+//   create an ExoRelay instance using environment variables
+//   connects to ExoCom
+//   registers the given handlers
+func Bootstrap(messageHandlers exorelay.MessageHandlerMapping) {
+	config := exosocket.NewEnvConfig()
+	exoRelay := exorelay.NewExoRelay(config)
+	err := exoRelay.Connect()
+	if err != nil {
+		panic(fmt.Sprintf("Error connecting: %v", err))
```
kevgo: Do we want Exocom to panic if there is an error? Why not return with an error here?

go/exorelay/index.go
```diff
@@ -0,0 +1,143 @@
...
+	"sync"
+	"time"
+
+	"github.com/Originate/exocom/go/exosocket"
+	"github.com/Originate/exocom/go/structs"
+)
+
+// ExoRelay is the mid level API to Exocom
+type ExoRelay struct {
+	closed               bool
```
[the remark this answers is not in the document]
charlierudolph: removed

go/exorelay/index.go
charlierudolph: We don't want to expose all of exosocket's methods. Only a couple

go/exorelay/index.go
charlierudolph: Yes. I'll log that as a separate issue

charlierudolph: https://github.com/Originate/exocom/issues/275

go/exorelay/test-fixtures/register_handler/main.go
charlierudolph: Added above the main functions

go/exoservice/index.go
charlierudolph: I prefer this API as its very high level where the user doesn't want to have to care about calling a method that returns an error. This is only one step below a CLI.

charlierudolph:
@kevgo all comments addressed

go/exorelay/index.go
kevgo: Interesting. Maybe explain in a comment why we deviate from the standard Go way here, so that the next person doesn't refactor this to using composition?

PR #281, opened 2017-12-07
update service data structure

charlierudolph:
For Originate/exosphere#853


go/exocom/src/exocom/parse_service_data.go
```diff
@@ -0,0 +1,42 @@
...
+		}
+		if route.Receives == nil {
+			route.Receives = []string{}
+		}
+		if route.MessageTranslations == nil {
+			route.MessageTranslations = []translation.MessageTranslation{}
+		}
+		parsed[serviceRole] = route
+	}
+	if _, ok := parsed["security"]; ok {
```
hugobho: Isn't there a `GetHasSecurity()` func to do this?

charlierudolph: Sorry. Missed this. I'll still apply this

